PR #1, opened 2016-10-31
PHPCS Configuration management

frenck:
Initial commit of a version which works by changing PHP_CodeSniffer's configuration


src/Plugin.php
```diff
@@ -11,22 +11,222 @@
...
-class Plugin implements PluginInterface
+class Plugin implements PluginInterface, EventSubscriberInterface
 {
+
+    const PACKAGE_TYPE = 'phpcodesniffer-standard';
+
+    /**
+     * @var Composer
+     */
+    protected $composer;
```
Potherca: If you do not plan on having the `Plugin` class extended, these properties might as well be `private`.

src/Plugin.php
```diff
@@ -11,22 +11,222 @@
...
+    }
+
+    /**
+     * Entry point for post install and post update events
+     *
+     * @throws RuntimeException
+     * @throws LogicException
+     * @throws ProcessFailedException
+     */
+    public function onScriptPost()
```
Potherca: The name `onScriptPost` seems unclear to me. I would prefer a `fooEventHandler` or `fooHandler` style name.

src/Plugin.php
```diff
@@ -11,22 +11,222 @@
...
+    {
+        // Ensure PHP_CodeSniffer is installed
+        if ($this->isPHPCodeSnifferInstalled() === false) {
+            return;
```
Potherca: False language construct. It would be better to invert the check (`!== false` or `=== true`) and move the statement below in to the `if` body. It might even be possible to collapse both `if`'s into one.

src/Plugin.php
```diff
@@ -11,22 +11,222 @@
...
+    /**
+     * Load all paths from PHP_CodeSniffer into an array
+     *
+     * @throws RuntimeException
+     * @throws LogicException
+     * @throws ProcessFailedException
+     */
+    private function loadInstalledPaths()
+    {
+        $phpcs = new Process($this->phpCodeSnifferBin . ' --config-show installed_paths');
```
Potherca: Instead of creating a new `Process` three times you might want to look into using the ProcessBuilder factory.

```
$this->builder = new ProcessBuilder();
$this->builder->setPrefix($this->phpCodeSnifferBin);
...
$this->builder
    ->setArguments(array('--list', '--file=archive.tar.gz'))
    ->getProcess()
    ->mustRun() /* or ->run()*/
```

src/Plugin.php
```diff
@@ -134,19 +139,18 @@ private function loadInstalledPaths()
...
-            $phpcsInstalledPaths = implode(',', $this->installedPaths);
-            $phpcs = new Process(
-                $this->phpCodeSnifferBin . ' --config-set installed_paths "' . $phpcsInstalledPaths . '"'
-            );
+        // By default we delete the installed paths
+        $arguments = ['--config-delete', 'installed_paths'];
+
+        // This changes in case we do have installed_paths
+        if (count($this->installedPaths) !== 0) {
+            $arguments = ['--config-set', 'installed_paths', implode(',', $this->installedPaths)];
```
Potherca: Is there a specific reason for not using `array_push ($argument, implode(',', $this->installedPaths));` instead of defining the entire array again?

PR #8, opened 2017-02-05
Show config actions and a result as Console output

### LINKED ISSUE #7: Install Plugin provides no feedback
## Problem/Motivation

After running a Composer install or update there is nothing in the command output to show what was changed (or not changed) by this installer plugin. 

## Expected behaviour

Composer will output the PHP CodeSniffer configuration key and value that it has set or changed.

## Actual behaviour

Nothing is shown to even indicate whether this Composer Plugin for PHP CodeSniffer was executed or not. If a PHP CodeSniffer configuration is changed, Composer does not say so.

## Steps to reproduce

1. Require this package with a supported standards package as per the README instructions.
1. Execute `composer install` or `composer update`

## Proposed changes

Use the Composer IO to write to output: 

1. Something to say the Plugin has started or is being run.
1. The config key and value to be set in PHP CodeSniffer.
1. (Optional) The results of the PHP CodeSniffer command to set a configuration key (ie. result from `vendor/bin/phpcs --config-set key value`)

christopher-hopper:
## Proposed Changes

1. Output to console config changes being made
1. When `-v` (verbose) flag is used, also output PHP CodeSniffer config command results

## Related Issues

Fixes #7

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -140,18 +140,29 @@ class Plugin implements PluginInterface, EventSubscriberInterface
      */
     private function saveInstalledPaths()
     {
-        // By default we delete the installed paths
-        $arguments = ['--config-delete', 'installed_paths'];
+        $configKey = 'installed_paths';
 
-        // This changes in case we do have installed_paths
+        // Check if we found installed_paths to set.
         if (count($this->installedPaths) !== 0) {
-            $arguments = ['--config-set', 'installed_paths', implode(',', $this->installedPaths)];
+            $paths = implode(',', $this->installedPaths);
+            $arguments = ['--config-set', $configKey, $paths];
+            $this->io->write(sprintf('PHP CodeSniffer Config <info>%s</info> <comment>set to</comment> <info>%s</info>', $configKey, $paths));
+        }
+        else {
+            // Delete the installed_paths if none were found.
+            $arguments = ['--config-delete', $configKey];
+            $this->io->write(sprintf('PHP CodeSniffer Config <info>%s</info> <comment>delete</comment>', $configKey));
         }
 
-        $this->processBuilder
+        $output = $this->processBuilder
             ->setArguments($arguments)
             ->getProcess()
-            ->mustRun();
+            ->mustRun()
+            ->getOutput()
+        ;
+        if ($this->io->isVerbose() && !empty($output)) {
+            $this->io->write(sprintf('<info>%s</info>', $output));
+        }
     }
 
     /**
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -147,22 +147,23 @@ class Plugin implements PluginInterface, EventSubscriberInterface
         if (count($this->installedPaths) !== 0) {
             $paths = implode(',', $this->installedPaths);
             $arguments = ['--config-set', $configKey, $paths];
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -140,18 +140,29 @@ private function loadInstalledPaths()
...
-        $arguments = ['--config-delete', 'installed_paths'];
+        $configKey = 'installed_paths';
 
-        // This changes in case we do have installed_paths
+        // Check if we found installed_paths to set.
         if (count($this->installedPaths) !== 0) {
-            $arguments = ['--config-set', 'installed_paths', implode(',', $this->installedPaths)];
+            $paths = implode(',', $this->installedPaths);
+            $arguments = ['--config-set', $configKey, $paths];
+            $this->io->write(sprintf('PHP CodeSniffer Config <info>%s</info> <comment>set to</comment> <info>%s</info>', $configKey, $paths));
```
Potherca: As a matter of cleanup, I would suggest assigning the message to a separate variable (instead of calling `$this->io->write()`) and placing the call to io->write outside of the if/else.

src/Plugin.php
```diff
@@ -141,18 +141,29 @@ private function loadInstalledPaths()
      */
     private function saveInstalledPaths()
     {
-        // By default we delete the installed paths
-        $arguments = ['--config-delete', 'installed_paths'];
+        $configKey = 'installed_paths';
```
frenck: This `configKey` (`installed_paths`) is used multiple times in this class / Plugin (e.g. line 121).

I would really like to keep a single style. Right now it used as a variable here, while being used directly elsewhere. I would suggest on moving it into a class constant.

src/Plugin.php
```diff
@@ -141,18 +141,29 @@ private function loadInstalledPaths()
...
 
-        // This changes in case we do have installed_paths
+        // Check if we found installed_paths to set.
         if (count($this->installedPaths) !== 0) {
-            $arguments = ['--config-set', 'installed_paths', implode(',', $this->installedPaths)];
+            $paths = implode(',', $this->installedPaths);
+            $arguments = ['--config-set', $configKey, $paths];
+            $this->io->write(sprintf('PHP CodeSniffer Config <info>%s</info> <comment>set to</comment> <info>%s</info>', $configKey, $paths));
+        }
+        else {
```
frenck: Violates [PRS-2: 5.1. if, elseif, else](http://www.php-fig.org/psr/psr-2/#if-elseif-else).
Please correct and place `else` on the same line as the closing bracket of the above if statement.

christopher-hopper:
Thanks for the review @frenck @Potherca -- I've pushed updates addressing your suggestions. Love your work.

I was a little unsure about the constant name. Do I go for the better readability, or more brevity. Landed somewhere in the middle I hope. *Naming things*: so difficult sometimes.

Potherca:
@christopher-hopper To (mis)quote [a poem by T.S. Eliot][1]:

>  The Naming of <del>Cats</del> <ins>a constant</ins> is a difficult matter,
> It isn't just one of your holiday games;
> You may think at first I'm as mad as a hatter
> When I tell you, a <del>cat</del> <ins>a constant</ins> must have THREE DIFFERENT NAMES.

😁

[1]: https://en.wikipedia.org/wiki/The_Naming_of_Cats

Issue #20, opened 2017-03-09
Check the root package for sniffs to install


Potherca:
## Problem/Motivation

As described in #19, the installer does not look in the root package. As it is completely viable for root packages to contain a code sniff, the installer should also check the root package.

## Expected behaviour / Actual behaviour / Steps to reproduce

_Not applicable._

## Proposed changes

A separate step needs to be implemented that checks the root package to see if it contains any sniffs that (also) need installing.

frenck:
Added this feature in PR #25 and is waiting code review.

frenck:
Merged into the master branch and will in the next release.

PR #25, opened 2017-05-09
Added support for having coding standards in the root package

frenck:
## Proposed Changes

This change adds support for having PHP_CodeSniffer standards in the root package.

## Related Issues

Implements: #20 Check the root package for sniffs to install

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -251,8 +251,8 @@ class Plugin implements PluginInterface, EventSubscriberInterface
     }
 
     /**
-     * Check all installed packages against the installed paths from
-     * PHP_CodeSniffer and add the missing ones.
+     * Check all installed packages (including the root package) against
+     * the installed paths from PHP_CodeSniffer and add the missing ones.
      *
      * @return bool True if changes where made, false otherwise
      *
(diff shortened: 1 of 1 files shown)


frenck:
Travis tests fails, enabled by @Potherca, but not configured it yet.
Therefore the failed Travis builds can be ignored for now.

src/Plugin.php
```diff
@@ -261,23 +261,26 @@ private function cleanInstalledPaths()
     private function updateInstalledPaths()
     {
         $changes = false;
-        $codingStandardPackages = $this->getPHPCodingStandardPackages();
 
+        $searchPaths = [getcwd()];
```
Potherca: For completeness sake: although usually discouraged using `getcwd` (and consorts) is the "Composer Way" of doing things.

frenck:
Just thought of an edge case that might cause troubles with this solution:
- This plugin is installed globally
- Current working directory is a repository containing a coding standard
- Running a composer global install/update command

This might cause the installation of the local coding standard into the global scope.

Will check on this tomorrow.

frenck:
Check my above concerns and concluded this won't be an issue since the Composer GlobalCommand will change the working directory _before_ actually running Composer.

See: https://github.com/composer/composer/blob/9e7d1a4f9ed34466a05ac1356e97a98686f7da67/src/Composer/Command/GlobalCommand.php#L77

Issue #26, opened 2017-05-10
Add support for code standards in root of repository for PHP_CodeSniffer 3.x


frenck:
## Problem/Motivation

PHP_CodeSniffer requires the coding standards to be in a directory, this behaviour changed in 3.x.

"The installed_paths config setting can now point directly to a standard. Previously, it had to always point to the directory in which the standard lives"

See the release notes:
https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.0.0a1

This plugin does not support this yet.

## Expected behaviour

Allowing a repository with a coding standard in the root of is project to work, when PHP_CodeSniffer 3.x or newer is used.

## Actual behaviour

The coding standard won't be installed by this plugin.

## Proposed changes

Change the coding standards search depth based on the PHP_CodeSniffer version used.

frenck:
Added in #30, closing issue.

PR #30, opened 2017-05-10
Support for coding standard in the root repository for PHP_CodeSniffer v3.x

frenck:
## Proposed Changes

Changes the search depth to start from 0, when PHP_CodeSniffer is version 3.x or newer. This also points the path of `installed_paths` for this PHP_CodeSniffer version directly to the coding standard instead of the parent directory.

## Related Issues

#26 Add support for code standards in root of repository for PHP_CodeSniffer 3.x

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -15,6 +15,7 @@ use Composer\EventDispatcher\EventSubscriberInterface;
 use Composer\IO\IOInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\PackageInterface;
+use Composer\Package\RootpackageInterface;
 use Composer\Plugin\PluginInterface;
 use Composer\Script\Event;
 use Composer\Script\ScriptEvents;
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -335,15 +346,17 @@ function (PackageInterface $package) {
     /**
      * Searches for the installed PHP_CodeSniffer Composer package
      *
+     * @param string|\Composer\Semver\Constraint\ConstraintInterface $versionConstraint to match against
```
Potherca: Missing `null` typehint.

src/Plugin.php
```diff
@@ -361,11 +374,13 @@ private function getPHPCodeSnifferInstallPath()
     /**
      * Simple check if PHP_CodeSniffer is installed.
      *
+     * @param string|\Composer\Semver\Constraint\ConstraintInterface $versionConstraint to match against
```
Potherca: Missing `null` typehint.

frenck:
Thank you @Potherca, updated the doc blocks.

PR #34, opened 2017-08-01
Fixes #33. Changes the way the installed_paths are set.

### LINKED ISSUE #33: Incorrect relative paths for WPCS
## Problem/Motivation

Possibly related to #32.

With the following in my composer:
```json
"dealerdirect/phpcodesniffer-composer-installer": "^0.4",
"wimg/php-compatibility": "dev-feature/prevent-conflicts-with-other-standards",
"wp-coding-standards/wpcs": "dev-develop"
```
the installed_paths phpcs config value is wrong.

Composer resolves this as using PHPCS 3.0.2, and the relative branches of WPCS and PHPCompatibility, both of which support PHPCS 3.*.

## Expected behaviour

I expect the paths to be correct. With 0.3 of this plugin. I get:

```php
<?php
 $phpCodeSnifferConfig = array (
  'installed_paths' => '/Users/gary/Local Sites/incipio/app/public/wp-content/themes/incipio/vendor/wp-coding-standards/wpcs,/Users/gary/Local Sites/incipio/app/public/wp-content/themes/incipio/vendor/wimg/php-compatibility',
)
?>
```

`phpcs -i` shows all of the standards correctly available, and `phpcs` runs correctly against the `phpcs.xml.dist` in my project.

## Actual behaviour

With 0.4 of the plugin, I get:

```
<?php
 $phpCodeSnifferConfig = array (
  'installed_paths' => '../../wp-coding-standards/wpcs/WordPress/,../../wp-coding-standards/wpcs/WordPress-Core/,../../wp-coding-standards/wpcs/WordPress-Docs/,../../wp-coding-standards/wpcs/WordPress-Extra/,../../wp-coding-standards/wpcs/WordPress-VIP/,../../wimg/php-compatibility/PHPCompatibility/',
)
?>
```

`phpcs -i` still does show the right standards are available, **but**, `phpcs` does not run correctly. Depending on the first called ruleset in `phpcs.xml.dist`, I get an error the form:

```
ERROR: Referenced sniff "WordPress" does not exist
```

This also happens if `phpcs . --standard=WordPress` is called.

Interestingly, calling `phpcs . --standard=WordPress-Core` does appear to run.

I'm not sure where the issue lies exactly, but @jrfnl says:
> PHPCS 3.x is supposed to support standards as a top-level folder (as in how PHPCompatibility used to be layed out and how DealerDirect registered the paths now), but that looks to be buggy based on your findings.
>
> I can run some tests with that and will report my finding to PHPCS, but basically as long as the `installed_paths` are registered in the PHPCS 2.x way (which is also still the recommendation for PHPCS 3.x), things should be fine.

frenck:
## Proposed Changes

Changes the behavior of registering installed_paths in PHP_CodeSniffer 3.x. It will now work in the same way as was required for PHP_CodeSniffer 2.x. It will record the directory ABOVE the directory containing the coding standard. The only exception is when the root project itself provides the standard in the root of its project.

## Related Issues

> [#33 Incorrect relative paths for WPCS](#33)

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -286,24 +286,25 @@ class Plugin implements PluginInterface, EventSubscriberInterface
             ->in($searchPaths);
 
         // Only version 3.x and higher has support for having coding standard in the root of the directory.
-        $allowCodingStandardsInRoot = $this->isPHPCodeSnifferInstalled('>= 3.0.0');
-
-        if ($allowCodingStandardsInRoot !== true) {
+        if ($this->isPHPCodeSnifferInstalled('>= 3.0.0') !== true) {
             $finder->depth('>= 1');
         }
 
+        // Process each found possible ruleset.
         foreach ($finder as $ruleset) {
             $standardsPath = $ruleset->getPath();
 
-            if ($allowCodingStandardsInRoot === false) {
+            // Pick the directory above the directory containing the standard, unless this is the project root.
+            if ($standardsPath !== getcwd()) {
                 $standardsPath = dirname($standardsPath);
             }
 
-            // Use relative paths for local project repositories
+            // Use relative paths for local project repositories.
             if ($this->isRunningGlobally() === false) {
                 $standardsPath = $this->getRelativePath($standardsPath);
             }
 
+            // De-duplicate and add when directory is not configured.
             if (in_array($standardsPath, $this->installedPaths, true) === false) {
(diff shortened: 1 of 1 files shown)


brammittendorff-dd:
Looks sane, you got my approve!

frenck:
Thanks for reviewing @brammittendorff-dd!

PR #38, opened 2017-09-08
Local PHPCS

GaryJones:
This plugin already has PHP_CodeSniffer as a dependency, but the project itself should also be checked against code standards.

While PSR-2 is already included, it would be useful to also check that the code supports down to the PHP 5.3 version specified in the Travis CI config file.

The inclusion of the `script` is so that those developing this repo can easily register the PHPCompatibility standard.

This is a prelude to #39.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
     "squizlabs/php_codesniffer": "*"
   },
   "require-dev": {
-    "composer/composer": "*"
+    "composer/composer": "*",
+    "wimg/php-compatibility": "^8.0"
   },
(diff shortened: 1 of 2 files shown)


Potherca:
Let me begin by stating that I am not opposed to this. It makes sense and if there are no major trade-offs I don't see any reason why not to merge this.

I don't have time to discuss this with @frenck until after the weekend, so we'll get back to you in the coming week.

frenck:
Approved, will update VersionEye to allow merging.

PR #39, opened 2017-09-08
CS: Add PHP 5.3 compatibility

GaryJones:
Logically follows on from #38, though isn't strictly dependent on it being merged.

The use of short syntax for arrays was the only thing stopping this package from being compatible with PHP 5.3 as well. PHP 5.3 is specified in the Travis CI config. Allowing it to be used in setups that require PHP 5.3 makes it usable for more authors (such as those in the WordPress community).

Composer also requires PHP 5.3.2 to run, so it makes sense to ensure it works with that too.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
     "source": "https://github.com/dealerdirect/phpcodesniffer-composer-installer"
   },
   "require": {
+    "php": "^5.3|^7",
     "composer-plugin-api": "^1.0",
     "squizlabs/php_codesniffer": "*"
   },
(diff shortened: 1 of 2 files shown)


Potherca:
Since this package does not use any explicit PHP 7.0 or 7.1 features, I think it would be sensible to follow the version constraint set by Composer.

GaryJones:
@frenck  Any feedback on this?

Here's an example of a [failing Travis build](https://travis-ci.org/wp-cli/wp-cli/jobs/275799214) which is now stopping [PRs](https://github.com/wp-cli/wp-cli/pull/4339) from being merged.

Potherca:
Unlike the VersionEye check that failed in PR-38, this check fails because of License whitelist issues.

screenshot from 2017-09-18 09-40-37

I've added the PHP license to the whitelist. Triggering the check again should have it pass the build.

PR #46, opened 2018-02-01
Adds the ability to set the max depth from the composer.json file

### LINKED ISSUE #45: Scan depth as parameter
## Problem/Motivation
This is feature request.

Current scan depth is set to less then 4. This may not work for cases when `ruleset.xml` is provided in the package that contains multiple rulesets - such package would use additional parent directory to hold multiple rulesets.

Exposing scan depth to be extracted from `extra` section would allow more granular  per-project config.

Potherca:
## Proposed Changes

This MR add the ability to set the max-depth of the finder from the "extra" section of the `composer.json` file. 

### Key

The name of the key is `phpcodesniffer-max-depth`. This can be changed if a better suggestion is made.

### Value

The value has the constraint that:
- It MUST be an integer (not a numeric string)
- Not smaller or equal to the minimum level (which is `0` or `1` depending on the version of PHPCS installed)
- Not so large it is converter to scientific notation (which registers as a float) as this triggers an exception from the Finder class.

If the constraint is not met an exception is thrown informing the user.

## Related Issues

Fixes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ in the Composer manual][composer-manual-scripts].
 
 ### Changing the Coding Standards search depth
 
-By default, this plugin searches up for Coding Standards up to four directories
+By default, this plugin searches up for Coding Standards up to three directories
 deep. In most cases, this should be sufficient. However, this plugin allows
 you to customize the search depth setting if needed.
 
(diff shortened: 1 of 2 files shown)


frenck:
Looks good 👍 
Missing docs...?

frenck:
@Potherca Should we move forward with this?

frenck:
Rebased.

frenck:
Changed it a little @Potherca...
Tested, documented and ready for merge -> I'll go ahead.

PR #51, opened 2018-07-11
Move loadInstalledPaths from init to onDependenciesChangedEvent

### LINKED ISSUE #49: The composer plugin implementation seems to be breaking the composer lifecycle
## Problem

The `dealerdirect/phpcodesniffer-composer-installer` version `0.4.4` (currently latest version) is probably  breaking composer lifecycle.

Here is a link to composer issue: https://github.com/composer/composer/issues/7298#issuecomment-393796348 where the Jordi Boggiano (the author of composer) and I have come to a conclusion that there may be a bug in this package.

Please, read the issue https://github.com/composer/composer/issues/7298, mainly the comment https://github.com/composer/composer/issues/7298#issuecomment-393796348, it contains all the details I was able to collect.

## Expected behaviour

The `composer update` or `composer require` should not crash in the custom installer implemented by this package while trying to use autoloader that is not yet regenerated and is in a stale state (not updated yet).

gapple:
## Proposed Changes

The plugin executes `phpcs` when initialized, which may cause it to include the autoloader while it's in an inconsistent state.

For example, if a package that includes a static file is removed during an update (e.g. a compat library that provides global PHP functions):
1. Composer removes the outdated package from the file system
2. The phpcodesniffer-composer-installer plugin is initialized and attempts to use the autoloader files on disk
3. The outdated autoloader files on disk require the file from the removed package, and the composer update / install fails with a fatal error.

This PR moves the initialization of `installedPaths` to the method which responds to the `POST_INSTALL_CMD` and `POST_UPDATES_CMD` events, so that it is only initialized just before it's actually needed.

## Related Issues

Fixes #49

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ before_install:
 
 install:
   - '[[ -f "${HOME}/bin/phpcs.phar" ]] || curl -L -o "${HOME}/bin/phpcs.phar" https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar'
-  - '[[ -f "${HOME}/bin/security-checker.phar" ]] || curl -L -o "${HOME}/bin/security-checker.phar" http://get.sensiolabs.org/security-checker.phar'
+  - '[[ -f "${HOME}/bin/security-checker-v4.1.5.phar" ]] || curl -L -o "${HOME}/bin/security-checker-v4.1.5.phar" http://get.sensiolabs.org/security-checker-v4.1.5.phar'
   - npm install -g jsonlint
 
 script:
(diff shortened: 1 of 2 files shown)


Potherca:
The failing build on Travis is caused by an issue with the security-checker.

It uses code that is only PHP7.1+ compatible.

gapple:
I've tagged security checker to the last release that supported PHP 5, and the Travis build is now successful.

Would it be better (and possible) to manage the version of security-checker via `require-dev` in `composer.json`?

Potherca:
I would be all for it but that is up to @frenck as I am no longer part of the organisation.

(I could ask @frenck to add me as a contributor... then I could make more contributions).

frenck:
This seems to be a sane fix for the issue at hand. I will test this one asap.

frenck:
@gapple About the security checker as a dependency, I'm kinda in the middle of this.
IMHO, security package checking is part of a CI and not part of a development workflow (e.g., GitHub is now moving towards providing that kind of things).

So I'm fine with both solutions tbh.

PR #56, opened 2018-09-07
README: replace maintenance badge

jrfnl:
## Proposed Changes

The `year` part of the Shields.io `maintained` badge needs to be updated with the current year for it not to show as "No" (unmaintained/abandoned).

I'd like to suggest using a different badge - "Last commit" - instead which doesn't need manual updating each year and still gives a reasonable indication of the maintenance status of the project.

## Related Issues

* https://shields.io/#/examples/other


frenck:
Agreed, thank you for the suggestion AND the PR.

PR #80, opened 2019-01-21
Respect PHP version used by Composer and provide better feedback on failure

### LINKED ISSUE #79: Composer PHP version appears not to be respected
## Problem/Motivation

Ok, so this is a fairly complicated one to explain, so please bear with me.

First off, I'm not sure if this is a Windows specific problem or a wider problem, but I have been able to consistently reproduce the issue on Windows.

```bash
> "vendor/bin/phpcs"
```

By default, if you call a script in the `vendor/bin` directory, the system default PHP version will be used. 
**Important**: This may not be the same PHP version as is used by Composer!

If you want commands to use the same PHP version as Composer uses, you need to define them as `scripts` and prefix them with `@php`.
See: https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts

```json
  "scripts": {
    "check-cs": "@php \"vendor/bin/phpcs\""
  }
```

For this to work on Windows with PHPCS (though probably the same goes for other tools), you can't point to the script in the `vendor/bin` directory as Windows will just print out that script, so you need to point to the actual script in the dependency.
See: https://github.com/composer/composer/issues/7645

```json
  "scripts": {
    "check-cs": "@php ./vendor/squizlabs/php_codesniffer/bin/phpcs"
  }
```

Now, given a system default PHP version of 5.4.x and a PHP used by Composer of 7.2.x and having PHPUnit as a project dependency, the DealerDirect PHPCS plugin will _appear_ to work correctly, but in reality will fail to work and the `CodeSniffer.conf` file is not created or updated.

The `PHPUnit` dependency is arbitrary, that is just the dependency through which I discovered the bug.
The point is to have a dependency which has non-cross-version compatible PHP code and which will autoload at least one file from the dependency.

I've tested this both with v 0.4.4 as well as 0.5.0.
In 0.5.0, this will fail silently.
In 0.4.4, this will fail with error info:
```
Writing lock file
Generating autoload files
PHP CodeSniffer Config installed_paths set to ../../phpcompatibility/php-compatibility/,../../wp-cod
ing-standards/wpcs/


  [Symfony\Component\Process\Exception\ProcessFailedException]
  The command ""path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/bin/phpcs" "--
  config-set" "installed_paths" "../../phpcompatibility/php-compatibility/,../../wp-coding-standa
  rds/wpcs/"" failed.

  Exit Code: 255(Unknown error)

  Working directory: path/to/Dealerdirect-PHPUnit-Conflict-POC

  Output:
  ================
  Parse error: syntax error, unexpected 'function' (T_FUNCTION), expecting identifier (T_STRING)
  or \\ (T_NS_SEPARATOR) in path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/my
  clabs/deep-copy/src/DeepCopy/deep_copy.php on line 5

  Call Stack:
      0.0010     129408   1. {main}() path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/squizlabs/php_codesniffer/bin/phpcs:0
      0.0040     172544   2. PHP_CodeSniffer\Autoload::load() path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/squizlabs/php_codesniffer/bin/phpcs:17
      0.0050     174312   3. include('path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/autoload.php') path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/squizlabs/php_codesniffer/autoload.php:79
      0.0070     186384   4. ComposerAutoloaderInit6a82c0ae3a6b11bce995085b1e490fcd::getLoader()    path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/autoload.php:7
      0.0170     389304   5. composerRequire6a82c0ae3a6b11bce995085b1e490fcd() path/to/Dealerdirect-PHPUnit-Conflict-POC/vendor/composer/autoload_real.php:56

  Error Output:
  ================


install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [-
-no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-a
utoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<package
s>]...
```

Based on the 0.4.4. output, it seems that the system default PHP version is used as some point - not the PHP version used by Composer - as this is a PHP 5.4 parse error for PHP 7 code.

While I have to say that it is bewildering to me that the referenced - completely unrelated - file would be loaded at this point in the first place (I'd guess that's a misconfigured autoload section in that dependency), it highlights that the default PHP version is used to set the `installed_paths` instead of the Composer version.

IMO, a Composer plugin should always respect the PHP version on which Composer is being run, which is why I'm reporting this here as a bug.

System info:
* Windows 7
* Composer 1.8.0
* Tested with version 0.4.4 and 0.5.0 of this plugin.

## Expected behaviour

That the PHPCS `installed_path` is set properly.

## Actual behaviour

The plugin echo's to the screen that the `installed_path` has been set, but in reality it hasn't been.

## Steps to reproduce

I've set up a minimal Proof of Conflict repo here: https://github.com/jrfnl/Dealerdirect-PHPUnit-Conflict-POC

To reproduce, you will need a Windows machine, though it may fail other OS-es as well.
* Set the default PHP version to PHP 5.4.x (i.e. set the various `PATH`s in Windows to point to a PHP 5.4.x install.)
* Edit the `composer.bat` file to point PHP to a more recent version.
* Clone the repo and check out the `master` branch.
* Run `composer install` and then run `phpcs-i`. All should be fine and it will display `The installed coding standards are MySource, PEAR, PSR1, PSR12, PSR2, Squiz, Zend, PHPCompatibility, WordPress, WordPress-Core, WordPress-Docs and WordPress-Extra`.
* Throw away `composer.lock` and the `vendor` directory.
* Check out the `POC` branch.
* Run `composer install` and then run `phpcs-i`. Now it will display `The installed coding standards are MySource, PEAR, PSR1, PSR12, PSR2, Squiz and Zend`.

## Proposed changes

I haven't dug into the plugin code yet to see what would need to be changed, but I'm guessing that if there are `exec()` commands and such, that will be where problem lies.

jrfnl:
### Improve feedback to user on failure to set paths

Related to #79.

When the setting of `installed_paths` fails, show an error instead of giving the impression that the setting of the `installed_paths` succeeded.

### Use the PHP version used by Composer

Possibly not the best way to do this, but it does work, so consider this a proof of concept, if not the solution.

The `getPhpExecCommand()` method in the `Composer\EventDispatcher\EventDispatcher` class basically does what is needed: find the PHP executable used by Composer and turn it into an executable command.

Unfortunately, that method is `protected`, making it difficult to call that method from within the plugin.

This now copies that method into the plugin and uses it to create the command to pass on to the `ProcessExecutor`.

The copied method has some minor modifications to comply with the coding standards used within this project.

Based on the proof of conflict referenced in #79, I can confirm that this would solve the issue.

## Related Issues

Fixes #79

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -239,12 +239,20 @@ class Plugin implements PluginInterface, EventSubscriberInterface
             );
         }
 
+        // Prepare message in case of failure
+        $failMessage = sprintf(
+            'Failed to set PHP CodeSniffer <info>%s</info> Config',
+            self::PHPCS_CONFIG_KEY
+        );
+
+        // Okay, lets rock! 🤘
+        $command = vsprintf('%s ./bin/phpcs %s', array(
+            'php executable' => $this->getPhpExecCommand(),
+            'arguments' => implode(' ', $arguments)
+        ));
+
         $exitCode = $this->processExecutor->execute(
-            sprintf(
-                '%s ./bin/phpcs %s',
-                $this->getPhpExecCommand(),
-                implode(' ', $arguments)
-            ),
+            $command,
             $configResult,
             $this->getPHPCodeSnifferInstallPath()
         );
@@ -252,10 +260,6 @@ class Plugin implements PluginInterface, EventSubscriberInterface
         if ($exitCode === 0) {
             $this->io->write($configMessage);
         } else {
-            $failMessage = sprintf(
-                'Failed to set PHP CodeSniffer <info>%s</info> Config',
-                self::PHPCS_CONFIG_KEY
-            );
             $this->io->write($failMessage);
         }
 
@@ -272,18 +276,25 @@ class Plugin implements PluginInterface, EventSubscriberInterface
     protected function getPhpExecCommand()
     {
         $finder = new PhpExecutableFinder();
+
         $phpPath = $finder->find(false);
-        if (!$phpPath) {
+
+        if ($phpPath === false) {
             throw new \RuntimeException('Failed to locate PHP binary to execute ' . $phpPath);
         }
-        $phpArgs = $finder->findArguments();
-        $phpArgs = $phpArgs ? ' ' . implode(' ', $phpArgs) : '';
 
-        $command  = ProcessExecutor::escape($phpPath);
-        $command .= $phpArgs;
-        $command .= ' -d allow_url_fopen=' . ProcessExecutor::escape(ini_get('allow_url_fopen'));
-        $command .= ' -d disable_functions=' . ProcessExecutor::escape(ini_get('disable_functions'));
-        $command .= ' -d memory_limit=' . ProcessExecutor::escape(ini_get('memory_limit'));
+        $phpArgs = $finder->findArguments();
+        $phpArgs = $phpArgs
+            ? ' ' . implode(' ', $phpArgs)
+            : ''
+        ;
+
+        $command  = ProcessExecutor::escape($phpPath) .
+            $phpArgs .
+            ' -d allow_url_fopen=' . ProcessExecutor::escape(ini_get('allow_url_fopen')) .
+            ' -d disable_functions=' . ProcessExecutor::escape(ini_get('disable_functions')) .
+            ' -d memory_limit=' . ProcessExecutor::escape(ini_get('memory_limit'))
+        ;
 
         return $command;
     }
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -25,6 +25,7 @@ use Symfony\Component\Finder\Finder;
 use Symfony\Component\Process\Exception\LogicException;
 use Symfony\Component\Process\Exception\ProcessFailedException;
 use Symfony\Component\Process\Exception\RuntimeException;
+use Symfony\Component\Process\PhpExecutableFinder;
 
 /**
  * PHP_CodeSniffer standard installation manager.
@@ -240,11 +241,12 @@ class Plugin implements PluginInterface, EventSubscriberInterface
 
         $exitCode = $this->processExecutor->execute(
             sprintf(
-                'phpcs %s',
+                '%s ./bin/phpcs %s',
+                $this->getPhpExecCommand(),
                 implode(' ', $arguments)
             ),
             $configResult,
-            $this->composer->getConfig()->get('bin-dir')
+            $this->getPHPCodeSnifferInstallPath()
         );
 
         if ($exitCode === 0) {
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -238,22 +239,55 @@ private function saveInstalledPaths()
...
+        $phpPath = $finder->find(false);
+        if (!$phpPath) {
+            throw new \RuntimeException('Failed to locate PHP binary to execute ' . $phpPath);
+        }
+        $phpArgs = $finder->findArguments();
+        $phpArgs = $phpArgs ? ' ' . implode(' ', $phpArgs) : '';
+
+        $command  = ProcessExecutor::escape($phpPath);
+        $command .= $phpArgs;
+        $command .= ' -d allow_url_fopen=' . ProcessExecutor::escape(ini_get('allow_url_fopen'));
```
exussum12: Just curious why are these needed ?

exussum12: ah never mind, its the comment above

exussum12:
Confirm this fixes it. Nice job :)

jrfnl:
@exussum12 Thanks for testing & confirming!

jrfnl:
Composer 1.9.0 has just been released and I noticed an entry in the changelog which may be useful to solve this issue, though that would mean that only Composer 1.9.0+ would be supported and I haven't run any tests yet to see if it is available to plugins as well as Composer scripts.

Anyway, just thought I'd leave a note about it here.

> Added PHP_BINARY as env var pointing to the PHP process when executing Composer scripts as shell scripts

Ref: https://github.com/composer/composer/releases/tag/1.9.0

exussum12:
Your fix basically uses that.

https://github.com/symfony/process/blob/master/PhpExecutableFinder.php#L36

The issue fixed in composer is for when the scripts called by composer are shell scripts (This is a direct php call) 

at least thats as i understand it - so this fix is still needed

jrfnl:
@exussum12 Thanks for looking into it. I hadn't had time yet. I was hoping it would remove the need for the function copy which is part of this PR.

jrfnl:
@frenck @Potherca Is there anything I can do to move this PR forward ?

Potherca:
For me, all that is needed is that I find a moment where I have both time, energy and a development machine available. As you may have noticed, it can take a while before the stars are thusly aligned...

I'm thinking that it might be beneficial at this point to add another developer to the list of maintainers as neither me nor @frenck can easily allocate time for this from our other responsibilities.

Potherca:
To get Travis passing MR #86 needs to be merged.

src/Plugin.php
```diff
@@ -272,18 +276,25 @@ private function saveInstalledPaths()
     protected function getPhpExecCommand()
```
jrfnl: FYI: I'd left the code in this function as close to the original as possible on purpose, with only minor CS changes to allow the build to pass.
That way it would be easier to compare the code with the source and update it if/when needed.

src/Plugin.php
```diff
@@ -239,23 +239,27 @@ private function saveInstalledPaths()
...
         }
 
+        // Prepare message in case of failure
+        $failMessage = sprintf(
+            'Failed to set PHP CodeSniffer <info>%s</info> Config',
+            self::PHPCS_CONFIG_KEY
+        );
+
+        // Okay, lets rock! 🤘
+        $command = vsprintf('%s ./bin/phpcs %s', [
```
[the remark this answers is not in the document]
Potherca: 🤦‍♂️

Potherca:
Awaiting MR #86, these changes can be seen to [pass the build](https://travis-ci.org/potherca-contrib/phpcodesniffer-composer-installer/builds/585262577), except for PHP5.3:

```
Failed to set PHP CodeSniffer installed_paths Config
```

jrfnl:
> The command "composer install-codestandards" exited with 0.

Should this exit code be `1` ?

Potherca:
Yup. I think that might be another bug.

mjrider:
Could you please rebase this on master, because the build issues have been resolved

jrfnl:
Rebased.

jrfnl:
I've rebased yet again and added an additional commit fixing the failure against PHPCS 2.x.

The build will currently fail on some code style issues which are addressed in PR #88. I'll rebase this PR again once that PR has been merged so you can see the build passing.

I'd really, really love to see this merged & released before the 1-year anniversary of the bug report.....

As per my comment in the ticket https://github.com/Dealerdirect/phpcodesniffer-composer-installer/issues/79#issuecomment-563053844, I've added another branch to the POC repo to make testing this fix even easier, but as can be seen in the comments from other people in this PR thread and based on my own tests, the fix works.

jrfnl:
Rebased & force pushed now #88 has been merged. We should now be able to see a 100% passing build.

Issue #82, opened 2019-06-18
Allow a string value for extra.phpcodesniffer-search-depth


TravisCarden:
It's impossible to set the `extra.phpcodesniffer-search-depth` config value from the command line because it strictly enforces an integer type on it, whereas `composer config` cannot set integer values (see https://github.com/composer/composer/issues/8195):

```
$ composer init -n
$ composer config extra.phpcodesniffer-search-depth 4
$ cat composer.json
{
    "require": {},
    "extra": {
        "phpcodesniffer-search-depth": "4"
    }
}
$ composer require --dev dealerdirect/phpcodesniffer-composer-installer
Using version ^0.5.0 for dealerdirect/phpcodesniffer-composer-installer
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 2 installs, 0 updates, 0 removals
  - Installing squizlabs/php_codesniffer (3.4.2): Loading from cache
  - Installing dealerdirect/phpcodesniffer-composer-installer (v0.5.0): Loading from cache
Writing lock file
Generating autoload files

Installation failed, reverting ./composer.json to its original content.


  [InvalidArgumentException]
  The value of "phpcodesniffer-search-depth" (in the composer.json "extra".section) must be an integer larger then 0, '4' given.


require [--dev] [--prefer-source] [--prefer-dist] [--no-progress] [--no-suggest] [--no-update] [--no-scripts] [--update-no-dev] [--update-with-dependencies] [--update-with-all-dependencies] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [--sort-packages] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--] [<packages>]...

```
I currently use the following fairly ugly kludge as a workaround:

```
sed -i'.bak' 's|"phpcodesniffer-search-depth": "4"|"phpcodesniffer-search-depth": 4|' composer.json && rm composer.json.bak
```

For its part, I've requested that Composer provide the ability to typecast data on the commandline (again, see https://github.com/composer/composer/issues/8195). On this side, it would be nice if `phpcodesniffer-composer-installer` would typecast string values to integers. Would the maintainers accept a pull request to that effect?

Potherca:
Hi! Thank you for reporting this issue!

We would definitely welcome such a contribution but (as you may have noticed from my late response) we don't have much time/energy to spend on this project currently.

[...]

PR #85, opened 2019-09-09
Add support for a string phpcodesniffer-search-depth config value set via `composer config`

TravisCarden:
## Proposed Changes

Add support for a string value of the `extra.phpcodesniffer-search-depth` config value so it can be set from the command line (`composer config`).

## Related Issues

https://github.com/Dealerdirect/phpcodesniffer-composer-installer/issues/82

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -441,9 +441,9 @@ class Plugin implements PluginInterface, EventSubscriberInterface
             $maxDepth = $extra[self::KEY_MAX_DEPTH];
             $minDepth = $this->getMinDepth();
 
-            if (is_int($maxDepth) === false     /* Must be an integer */
-                || $maxDepth <= $minDepth       /* Larger than the minimum */
-                || is_float($maxDepth) === true /* Within the boundaries of integer */
+            if ((string) (int) $maxDepth !== (string) $maxDepth /* Must be an integer or cleanly castable to one */
+                || $maxDepth <= $minDepth                       /* Larger than the minimum */
+                || is_float($maxDepth) === true                 /* Within the boundaries of integer */
             ) {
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -441,9 +441,9 @@ private function getMaxDepth()
             $maxDepth = $extra[self::KEY_MAX_DEPTH];
             $minDepth = $this->getMinDepth();
 
-            if (is_int($maxDepth) === false     /* Must be an integer */
-                || $maxDepth <= $minDepth       /* Larger than the minimum */
-                || is_float($maxDepth) === true /* Within the boundaries of integer */
+            if ((string) (int) $maxDepth !== (string) $maxDepth /* Must be an integer or cleanly castable to one */
```
Potherca: I've never seen this before... Is there any benefit of this over using [`is_numeric`](https://www.php.net/manual/en/function.is-numeric.php)?

TravisCarden: Yes, @Potherca. `is_numeric` would match floating point numbers, whereas we only want to match integers and string representations of them. There's no builtin function for that, as you know, but typecasting a string to an integer and back and then seeing whether it matches the original value is a common idiom for it. It's probably it's fairly intuitive once you've been introduced to it. :)

Potherca:
To get Travis passing MR #86 needs to be merged.

Potherca:
Awaiting MR #86, these changes can be seen to [pass the build in Travis](https://travis-ci.org/potherca-contrib/phpcodesniffer-composer-installer/builds/585267925).

mjrider:
Could you please rebase this on master, because the build issues have been resolved

TravisCarden:
Rebased, @mjrider.

PR #86, opened 2019-09-11
Update Travis file and fix build

Potherca:
## Proposed Changes

Currently the Travis build is breaking because the URL used by the security-checker no longer exists.
This MR updates that URL to the new home in order to fix the build.

It also adds a preview for PHP 7.4 and makes minor changes to improve the build step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,6 @@ php:
   - 7.1
   - 7.0
   - 5.6
-  - 5.5
-  - 5.4
-
-# This triggers builds to run on the new (faster) TravisCI infrastructure.
-# See: http://docs.travis-ci.com/user/workers/container-based-infrastructure/
-sudo: false
 
 cache:
   directories:
@@ -25,8 +19,14 @@ matrix:
   include:
     - php: 7.2
       env: SNIFF=1
+    - php: 5.5
+      # As the latest Debian does not support PHP 5.5 anymore, we need to force using 'trusty'.
+      dist: trusty
+    - php: 5.4
+      # As the latest Debian does not support PHP 5.4 anymore, we need to force using 'trusty'.
+      dist: trusty
     - php: 5.3
-      # As 'trusty' is not supporting PHP 5.2/5.3 anymore, we need to force using 'precise'.
+      # As the latest Debian does not support PHP 5.3 anymore, we need to force using 'precise'.
       dist: precise
 
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,4 +45,4 @@ script:
   - travis_wait composer install --no-interaction --no-progress --no-scripts --no-suggest --optimize-autoloader --prefer-dist --verbose
   - if [[ "$SNIFF" == "1" ]]; then composer install-codestandards; fi
   - if [[ "$SNIFF" == "1" ]]; then ./vendor/bin/phpcs; fi
-  - ./vendor/bin/security-checker -n security:check --end-point=http://security.sensiolabs.org/check_lock
+  - ./vendor/bin/security-checker -n security:check --end-point=http://security.symfony.com/check_lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: php
 
 php:
-  - 7.3
+  - 7.2
   - 7.1
   - 7.0
   - 5.6
@@ -17,7 +17,7 @@ env:
 
 matrix:
   include:
-    - php: 7.2
+    - php: 7.3
       env: SNIFF=1
     - php: 5.5
       # As the latest Debian does not support PHP 5.5 anymore, we need to force using 'trusty'.
@@ -39,10 +39,11 @@ install:
   - npm install -g jsonlint
 
 script:
-  - find . -type f -name "*.json" -print0 | xargs -0 -n1 jsonlint -q
-  - find . -type f -name "*.php" -print0 | xargs -0 -n1 php -l
-  - composer validate
+  - if [[ "$SNIFF" == "1" ]];then find . -type f -name "*.json" -print0 | xargs -0 -n1 jsonlint -q; fi
+  - if [[ "$SNIFF" == "1" ]];then find . -type f -name "*.php" -print0 | xargs -0 -n1 php -l; fi
+  - if [[ "$SNIFF" == "1" ]];then composer validate; fi
   - travis_wait composer install --no-interaction --no-progress --no-scripts --no-suggest --optimize-autoloader --prefer-dist --verbose
-  - if [[ "$SNIFF" == "1" ]]; then composer install-codestandards; fi
-  - if [[ "$SNIFF" == "1" ]]; then ./vendor/bin/phpcs; fi
-  - ./vendor/bin/security-checker -n security:check --end-point=http://security.symfony.com/check_lock
+  - composer install-codestandards
+  - ./vendor/bin/phpcs -i
+  - ./vendor/bin/phpcs --standard=PHPCompatibility --runtime-set testVersion "${TRAVIS_PHP_VERSION:0:3}" --exclude=PHPCompatibility.Upgrade.LowPHP ./src/
+  - if [[ "$SNIFF" == "1" ]];then ./vendor/bin/security-checker -n security:check --end-point=http://security.symfony.com/check_lock; fi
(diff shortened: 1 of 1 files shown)


Potherca:
@mjrider, can you review (and optionally merge) this, if @frenck is unavailable?

Potherca:
Looks like this fixes _part_ of the issue... PHP5.4 and PHP5.5 need Trusty to build.

However, changing the matrix to this:

```
matrix:
  include:
    - php: 7.2
      env: SNIFF=1
    - php: 5.5
      # As the latest Debian not supporting PHP 5.5 anymore, we need to force using 'trusty'.
      dist: trusty
    - php: 5.4
      # As the latest Debian not supporting PHP 5.4 anymore, we need to force using 'trusty'.
      dist: trusty
    - php: 5.3
      # As the latest Debian not supporting PHP 5.3 anymore, we need to force using 'precise'.
      dist: precise
```
did not yield the expected result (i.e. Trusty being used...)

jrfnl:
Oh and as you're making changes to the Travis script now anyway, why not add a build against `"7.4snapshot"`, possibly with an `allow_failures` for it too ?

Potherca:
@mjrider I'm all done. Can you please review/approve so these changes can be merged to master?

They are needed before I can update the other open MRs.

.travis.yml
```diff
@@ -39,10 +41,11 @@ install:
   - npm install -g jsonlint
 
 script:
-  - find . -type f -name "*.json" -print0 | xargs -0 -n1 jsonlint -q
-  - find . -type f -name "*.php" -print0 | xargs -0 -n1 php -l
-  - composer validate
+  - if [[ "$SNIFF" == "1" ]];then find . -type f -name "*.json" -print0 | xargs -0 -n1 jsonlint -q; fi
+  - if [[ "$SNIFF" == "1" ]];then find . -type f -name "*.php" -print0 | xargs -0 -n1 php -l; fi
```
jrfnl: Shouldn't PHP linting be done against every PHP version ?

.travis.yml
```diff
@@ -39,10 +41,11 @@ install:
...
+  - if [[ "$SNIFF" == "1" ]];then composer validate; fi
```
jrfnl: As there is no committed `composer.lock`, you may want to validate against both the highest as well as the lowest PHP version.

.travis.yml
```diff
@@ -39,10 +41,11 @@ install:
...
   - travis_wait composer install --no-interaction --no-progress --no-scripts --no-suggest --optimize-autoloader --prefer-dist --verbose
-  - if [[ "$SNIFF" == "1" ]]; then composer install-codestandards; fi
-  - if [[ "$SNIFF" == "1" ]]; then ./vendor/bin/phpcs; fi
-  - ./vendor/bin/security-checker -n security:check --end-point=http://security.sensiolabs.org/check_lock
+  - composer install-codestandards
+  - ./vendor/bin/phpcs -i
+  - ./vendor/bin/phpcs --standard=PHPCompatibility --runtime-set testVersion "${TRAVIS_PHP_VERSION:0:3}" --exclude=PHPCompatibility.Upgrade.LowPHP ./src/
```
jrfnl: The sniffs will return the same results cross-version (and tested for this in PHPCompatibility itself), so you're now effectively unit testing PHPCompatibility, which you don't need to do.

.travis.yml
```diff
@@ -39,10 +41,11 @@ install:
...
+  - if [[ "$SNIFF" == "1" ]];then ./vendor/bin/security-checker -n security:check --end-point=http://security.symfony.com/check_lock; fi
```
jrfnl: As there is no committed composer.lock, you may want to run the security check against both the highest as well as the lowest PHP version.

jrfnl:
> The command "composer install-codestandards" exited with 0.

Should this exit code be `1` ?

.travis.yml
```diff
@@ -39,10 +41,11 @@ install:
...
 script:
-  - find . -type f -name "*.json" -print0 | xargs -0 -n1 jsonlint -q
-  - find . -type f -name "*.php" -print0 | xargs -0 -n1 php -l
-  - composer validate
+  - if [[ "$SNIFF" == "1" ]];then find . -type f -name "*.json" -print0 | xargs -0 -n1 jsonlint -q; fi
+  - if [[ "$SNIFF" == "1" ]];then find . -type f -name "*.php" -print0 | xargs -0 -n1 php -l; fi
+  - if [[ "$SNIFF" == "1" ]];then composer validate; fi
   - travis_wait composer install --no-interaction --no-progress --no-scripts --no-suggest --optimize-autoloader --prefer-dist --verbose
-  - if [[ "$SNIFF" == "1" ]]; then composer install-codestandards; fi
-  - if [[ "$SNIFF" == "1" ]]; then ./vendor/bin/phpcs; fi
```
jrfnl: By removing this, you are removing the CS check on the code of the plugin itself. Was that intentional ?

.travis.yml
mjrider: iirc composer validate only checks of the composer files are valid, independent of php version

.travis.yml
mjrider: @jrfnl i don't follow, so can you elaborate please?

what this is doing afaik is running it phpcs in 'run in php version X' which should give different warnings/errors depending on when a featured was added or removed in php

.travis.yml
mjrider: personally, i don't care about the oldest version, because if there is an security issue there is a big change we can't fix it because the update on upstream requires a newer php version

.travis.yml
jrfnl: @mjrider This command (as-is) is running the PHPCompatibility sniffs against each PHP version and with a `testVersion` which is the same as the PHP version the command is run on.

PHPCompatibility, however, supports `testVersion` ranges, so, running it once on a high PHP version with a `testVersion` range will get you the same or even better results (some issue will only display when you want to support a range of PHP versions).

`./vendor/bin/phpcs --standard=PHPCompatibility --runtime-set testVersion 5.3- --exclude=PHPCompatibility.Upgrade.LowPHP ./src/`

Ref: https://github.com/PHPCompatibility/PHPCompatibility#sniffing-your-code-for-compatibility-with-specific-php-versions

Better yet, run the normal CS check (as mentioned below) and it includes the PHPCompatibility check with the correct `testVersion`.

See: https://github.com/Dealerdirect/phpcodesniffer-composer-installer/blob/master/phpcs.xml.dist#L15-L16

.travis.yml
jrfnl: Fair enough and as this project will never really be used stand-alone, but only ever as a dependency for other projects, maybe those projects should do that check instead anyway.

.travis.yml
mjrider: i see what you are doing,
by running it per every version, you get a specific build to fail. so when there is a change breaking php 5.6 but not 5.5 and 7.0 only the build for php 5.6 breaks, so without even reading the build output, you already know which php version is breaking.

with this pipeline, it is checking php 7.4 with allow failure,
so effectively we are running it for 5.3-7.3 and 7.4 but we do not care about it complaining.

imho it's a matter of taste if you take one way or the other. Personally i would run it for every version separately

.travis.yml
mjrider: Fixes by running it together with the PHPCompatibility standard

.travis.yml
jrfnl: > imho it's a matter of taste if you take one way or the other

It's not a matter of taste. It's a matter of understanding how this works. I clearly did not explain this well enough.

You can have separate runs with a different fixed `testVersion` setting, but...
1. The results will be **_less_ accurate** compared to running the sniffs once with a `testVersion` range.
    As I said above, some sniffs will only trigger when a combination of PHP versions needs to be supported.
    For example: PHP changed the default charset for some functions in PHP 5.4. If you only run with `5.3` or `5.4` as `testVersion` you will not receive an error for that, as when only that version needs to be supported, it doesn't cause a problem.
    However, if you run with a `testVersion` of `5.3-5.4`, you will receive an error.
2. While generally speaking PHPCompatibility will give the same results (given the same `testVersion`) across all PHP versions on which PHPCompatibility is supported, there are some tokenizer quirks which can't be worked-around.
    So for the _best_ results, it is strongly recommended to run on a high PHP version.

.travis.yml
jrfnl: Nope, sorry, you've made things worse.

PR #88, opened 2019-12-08
CS: fix compliance with PSR12

jrfnl:
## Proposed Changes

PHPCS 3.5.0+ contains a far more complete PSR12 ruleset.

1. The current code didn't fully comply with that ruleset. Some minor code tweaks fix that.
2. PSR12 demands constant visibility to be declared, but as this project still supports PHP 5.3 and constant visibility wasn't introduced until PHP 7.1, this is not desirable. Excluding that particular rule from within the ruleset fixes that.

## Related Issues

* https://github.com/squizlabs/php_codesniffer/releases

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -15,5 +15,9 @@
     <rule ref="PHPCompatibility"/>
     <config name="testVersion" value="5.3-"/>
 
-    <rule ref="PSR12"/>
+    <rule ref="PSR12">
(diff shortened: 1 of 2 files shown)


jrfnl:
@mjrider Thanks. Now we have a passing build again, I'll be sending in some more PRs to further improve the testing of the plugin. I'm sending them in staged as otherwise we'd have to fix conflicts all the time.

PR #97, opened 2020-01-17
Verify the `installed_paths` after save

jrfnl:
## Proposed Changes

As most problems reported are to do with paths not being set correctly, let's verify success within the plugin, independently of PHPCS.

To that end, a new function `verifySaveSuccess()` has been added.
This function takes the paths which were to be saved and compares them to the paths which are actually set in PHPCS after the save to determine whether or not the plugin was successful and changes the exit code if this is not the case.

Open questions:
* Should additional debug output be generated by the verification function when `verbose` mode is turned on ? Something like `Expected paths to be set to: %s, found: %s`.
    This could also come in handy to verify that all paths which should be set were found by the plugin.

At this moment I don't have a test case for this new functionality, however, I imagine it will come in useful with future bug reports.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -214,7 +214,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
             if (preg_match($regex, $output, $match) === 1) {
                 $phpcsInstalledPaths = str_replace(self::PHPCS_CONFIG_KEY . ': ', '', $match[0]);
                 $phpcsInstalledPaths = trim($phpcsInstalledPaths);
-    
+
                 if ($phpcsInstalledPaths !== '') {
                     $this->installedPaths = explode(',', $phpcsInstalledPaths);
                 }
@@ -278,6 +278,9 @@ class Plugin implements PluginInterface, EventSubscriberInterface
         );
 
         $exitCode = $this->processExecutor->execute($command, $configResult, $phpcsPath);
+        if ($exitCode === 0) {
+            $exitCode = $this->verifySaveSuccess();
+        }
 
         if ($exitCode === 0) {
             $this->io->write($configMessage);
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -288,6 +291,33 @@ private function saveInstalledPaths()
...
+        if ($expectedCount === 0) {
+            if (count($this->installedPaths) === 0) {
+                $exitCode = 0;
+            }
+        } else {
+            $registered = array_intersect($this->installedPaths, $expected);
+            if ($expectedCount === count($registered)) {
+                $exitCode = 0;
+            }
+        }
```
Potherca: This structure feel hard to read to me. It might be a matter of taste, but as this comparison basically boils down to this:
```
if (($a && $b) || (!$a && $c)){
    $exitCode = 0;
}
```
I would prefer something more explicit, like this:

```
        $registered = array_intersect($this->installedPaths, $expected);
        $installedCount = count($this->installedPaths)
        $registeredCount = count($registered)

        if (
            ($expectedCount === 0 && $installedCount === 0)
            || ($expectedCount > 0 && $expectedCount === $registeredCount)
        ) {
            $exitCode = 0;
        }
```

Potherca:
> Should additional debug output be generated by the verification function when verbose mode is turned on?

I know that goes against the principle of only reporting errors in output, but as this is in verbose mode, I would say **Yes, definitely.**

> I imagine it will come in useful with future bug reports.

Fully agree. Should make life easier for all.

src/Plugin.php
jrfnl: @Potherca That would mean the relatively expensive `array_intersect()` would be run, even when it is not needed (`$expectedCount === 0`).

jrfnl:
Oh oh oh ... while debugging why the build was failing, I've discovered a pretty interesting bug, which could well be the cause of some reported mystery issues. Win!

I'll work on a fix and send it in as a separate PR which will need to be merged before this PR as this PR will be dependent on it.

src/Plugin.php
Potherca: True, but I don't think we need to worry about optimisations too much at this point.
Readability of the code would be of greater concern as far as I'm concerned...

Potherca:
PR #98 has been merged.

jrfnl:
Thanks, I'll update this PR in a moment and will make it mergable if the build then passes.

src/Plugin.php
jrfnl: In that case, we don't need to _special case_ the `$expectedCount === 0` either.

Potherca: Good point.

jrfnl:
Marking this ready for review. I've updated the PR for:
- code layout
- verbose output on failure

jrfnl:
Just now thinking: this should probably also run `cleanInstalledPaths()` to safeguard against the comparison failing for Windows vs Linux slashes. I'll send in a new PR to add that.

jrfnl:
Oh second thought: no, as `$this->installedPaths` is only updated when the path doesn't exist, `cleanInstalledPaths()` doesn't update for slash differences.

Hmm.. let's see how we fair for now. If needs be, this can be revisited later.

PR #98, opened 2020-01-19
Bug fix: `loadInstalledPaths()` very very broken since PHPCS 3.1.0

jrfnl:
## Proposed Changes

Discovered while debugging the build failure of #97

The PHPCS `--config-show` command shows **all** config settings set, not just the `installed_paths` and since PHPCS 3.1.0, it also includes a line to indicate which configuration file is being used to help with debugging configuration issues.

Example:
```
Using config file: I:/path/to/project/vendor/squizlabs/php_codesniffer/CodeSniffer.conf

default_standard:  Squiz
installed_paths:   ../../phpcompatibility/php-compatibility
show_progress:     1
```

In PHPCS 2.0 - 3.0.1, if any [other configuration options](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options) were set, aside from `installed_paths`, the result of the `loadInstalledPaths()` command would be broken.

In PHPCS 3.1.0+, the result would always be broken, no matter what.

In effect this means that the `installed_paths` would (nearly) always be set as if the config was previously non-existent as the `cleanInstalledPaths()` method would empty out the array as the values contained therein would be invalid anyway.

It would be lovely if it were possible to add unit tests, but as we're talking about a `private` method setting a `private` property...

So, for now, to proof the bug and validate the fix, I've set up a "fake" unit test to demonstrate what the previous behaviour was and what the behaviour will be with this fix in place: https://3v4l.org/brrNT

I suspect that this may fix some mystery issues previously reported.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -46,6 +46,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
     const PACKAGE_NAME = 'squizlabs/php_codesniffer';
     const PACKAGE_TYPE = 'phpcodesniffer-standard';
 
+    const PHPCS_CONFIG_REGEX = '`%s:[^\r\n]+`';
     const PHPCS_CONFIG_KEY = 'installed_paths';
 
     /**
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -199,22 +200,29 @@ public function onDependenciesChangedEvent()
...
-            $this->processExecutor->execute(
-                sprintf(
-                    'phpcs --config-show %s',
-                    self::PHPCS_CONFIG_KEY
-                ),
-                $output,
-                $this->composer->getConfig()->get('bin-dir')
-            );
+        if ($this->isPHPCodeSnifferInstalled() !== true) {
+            return;
```
Potherca: Personally, I am rather opposed to the "early return" pattern. Besides being a way of hiding complexity, it is also a misrepresentation of what is actually happening in the flow.

In this specific instance it also has the negative side-effect of making the changeset a lot larger than it needs to be because of the change in whitespace.

src/Plugin.php
```diff
@@ -199,22 +200,29 @@ public function onDependenciesChangedEvent()
...
+            ),
+            $output,
+            $this->composer->getConfig()->get('bin-dir')
+        );
 
-            $phpcsInstalledPaths = str_replace(self::PHPCS_CONFIG_KEY . ': ', '', $output);
-            $phpcsInstalledPaths = trim($phpcsInstalledPaths);
+        $regex = sprintf(self::PHPCS_CONFIG_REGEX, self::PHPCS_CONFIG_KEY);
+        if (preg_match($regex, $output, $match) !== 1) {
+            return;
```
Potherca: I would rather have the code that is supposed to be run for the `true` scenario onside the `if` body, rather than (ab)using an early return.

src/Plugin.php
jrfnl: Done & done.

Most projects I contribute to have a preference for early return, I'm not that bothered either way.

Anyway, I've changed it around now.

Issue #103, opened 2020-01-25
Do not exit with code 1 on uninstall (--no-dev)


tobiasbaehr:
## Problem/Motivation

When dealerdirect/phpcodesniffer-composer-installer + package which has phpcs rules are uninstalled while `composer install --no-dev` the dealerdirect/phpcodesniffer-composer-installer exit with the exitcode 1, which breaks a CI.

## Expected behaviour

Do nothing in this case.

## Actual behaviour

```
$ composer install --no-interaction --optimize-autoloader --ansi --no-progress --no-dev
00:03
 No patches supplied.
 Loading composer repositories with package information
 Installing dependencies from lock file
 Package operations: 0 installs, 0 updates, 5 removals
   - Removing squizlabs/php_codesniffer (3.5.3)
 Deleting /builds/foobar/vendor/squizlabs/php_codesniffer - deleted
   - Removing drupal/stage_file_proxy (1.0.0-rc2)
 Deleting web/modules/contrib/stage_file_proxy - deleted
   - Removing drupal/devel (2.1.0)
 Deleting web/modules/contrib/devel - deleted
   - Removing drupal/coder (8.3.7)
   - Removing dealerdirect/phpcodesniffer-composer-installer (v0.6.0)
 Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.
 Package zendframework/zend-diactoros is abandoned, you should avoid using it. Use laminas/laminas-diactoros instead.
 Package zendframework/zend-escaper is abandoned, you should avoid using it. Use laminas/laminas-escaper instead.
 Package zendframework/zend-feed is abandoned, you should avoid using it. Use laminas/laminas-feed instead.
 Package zendframework/zend-stdlib is abandoned, you should avoid using it. Use laminas/laminas-stdlib instead.
 Generating optimized autoload files
 ERROR: Job failed: exit code 1
```

## Steps to reproduce

Run `composer install --no-interaction --optimize-autoloader --ansi --no-progress --no-dev -vv;echo $?` to remove the dev-deps and the see the exit code.

## Proposed changes

> (If you have a proposed change, workaround or fix, describe the rationale behind it)

## Environment

| Question               | Answer
| ------------------------| -------
| OS                                | Mac
| PHP version                       | 7.3.13
| Composer version                  | 1.9.2
| PHP_CodeSniffer version           | 3.5.3
| Dealerdirect PHPCS plugin version | v0.6.0
[...]

PR #104, opened 2020-01-26
Fix for issue #103

Potherca:
If the phpcodesniffer-composer-installer plugin is installed as a dev
requirement and it is then uninstalled as part of a "--no-dev" install, a bug
occurs. The bug that occurs is that the plugin complains that the  package
"squizlabs/php_codesniffer" is not installed without checking if the package
should be present.

## Proposed Changes

This commit adds a check to verify that this plugin is actually installed before
complaining about the missing package. If this plugin itself is removed, then
it should not complain about the missing package.

## Related Issues

#103

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -41,6 +41,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
         'The value of "%s" (in the composer.json "extra".section) must be an integer larger then %d, %s given.';
     const MESSAGE_NOT_INSTALLED = 'PHPCodeSniffer is not installed';
     const MESSAGE_NOTHING_TO_INSTALL = 'Nothing to install or update';
+    const MESSAGE_PLUGIN_UNINSTALLED = 'PHPCodeSniffer Composer Installer is uninstalled';
     const MESSAGE_RUNNING_INSTALLER = 'Running PHPCodeSniffer Composer Installer';
 
     const PACKAGE_NAME = 'squizlabs/php_codesniffer';
@@ -48,6 +49,8 @@ class Plugin implements PluginInterface, EventSubscriberInterface
 
     const PHPCS_CONFIG_REGEX = '`%s:[^\r\n]+`';
     const PHPCS_CONFIG_KEY = 'installed_paths';
+
+    const PLUGIN_NAME = 'dealerdirect/phpcodesniffer-composer-installer';
 
     /**
      * @var Composer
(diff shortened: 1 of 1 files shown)


src/Plugin.php
```diff
@@ -182,9 +185,26 @@ public function onDependenciesChangedEvent()
                 $io->write(sprintf('<info>%s</info>', self::MESSAGE_NOTHING_TO_INSTALL));
             }
         } else {
-            $exitCode = 1;
-            if ($isVerbose) {
-                $io->write(sprintf('<info>%s</info>', self::MESSAGE_NOT_INSTALLED));
+            $isDev = array_key_exists(self::PLUGIN_NAME, $this->composer->getPackage()->getDevRequires());
```
jrfnl: :point_up: This looks like it only takes the `dev` requires into account, but the issue isn't limited to `dev` requires.

Potherca: Fixed.

Potherca:
You are correct... The "Is this plugin installed?" check should always run. It doesn't matter if it is `require-dev` or `require`. I've updated my test scenarios and removed the `$isDev`.

I guess I was a bit monofocussed on the reported scenario. 🤷

Anyway, good catch! 👍

Regarding the "P.S."... I tend to always work from my own fork, for various reasons. 

1. I would rather not clutter the main repo with more branches than needed. If anyone clones the main repo, they get the other branches as well. So unless it is a large feature that more developers are working on, I would rather keep it out of the main repo.
2. I have a habit of keeping my (work) processes as uniform as possible. Since I can't  push to any repo of choice, I tend to just always work from my own fork. This also avoids issues when/if I no longer have access to the main repo. By working from my fork, there is no need to change (or even notice) when this happens.
3. I don't feel any reason why I _should_, to be honest... I've gotten so used to the distributed nature of git that the "one blessed repository" with a bunch of developer pushing to it feels like an anti-pattern. In FOSS this is less of an issue, but in a corporate setting I have (almost always) seen this lead to situations where people don't know if certain branches can be deleted or not. It creates overhead, so it feels like an anti-pattern to me.

jrfnl:
Thanks @Potherca for sorting this out.

I think I will still an issue in the Composer repo to ask about the principle of plugins running after they've been uninstalled. Depending on the answer there, we can iterate on this further.

---

Re: the "P.S.":

Thanks for the extensive answer. I was mostly wondering about it as it makes (manual) testing by other committers or even by casual watchers of the repo more awkward.

To be able to test this PR, as you will no doubt be aware, I had to add a new remote + add a `repositories` `vcs` directive to the `composer.json` for the test scenario I was using.
If the branch had been in this repo, neither would have been necessary.

As PR branches should "normally" be short-lived, I'm not too concerned about branches being added to a repo, especially as it is now possible for them to be auto-deleted on merge (setting in the repo settings).

I'm not trying to persuade you to change your workflow. Just wanted to give you my perspective.

PR #106, opened 2020-02-14
Allow installation on PHP 8

jrfnl:
## Proposed Changes

As it was, the plugin when used by external standards, would block them from being tested on PHP 8.

As all the dependencies in the `require` section allow for PHP 8, there is no reason not to allow it.

The only dependencies currently in use which do not allow for PHP 8 yet are in the `require-dev` section. See the below output.

This includes updating the Travis script to:
* Only do a full (`dev`) install when not on `nightly`.
* Ignore platform requirements on `nightly` as otherwise the `dev` requirements would still block installation.
* And as PHPCompatibility is in the `require-dev` section, we need to require it explicitly for `nightly` to make sure the integration test can run.

As the Travis build on PHP `nightly` passes, I'm semi-confident that all should be good, though confidence will grow once we add more unit/integration tests.

```bash
# composer prohibits php:8

composer/composer          1.9.3   requires  php (^5.3.2 || ^7.0)
composer/semver            1.5.1   requires  php (^5.3.2 || ^7.0)
seld/jsonlint              1.7.2   requires  php (^5.3 || ^7.0)
symfony/console            v4.4.4  requires  php (^7.1.3)
symfony/filesystem         v4.4.4  requires  php (^7.1.3)
symfony/finder             v4.4.4  requires  php (^7.1.3)
symfony/process            v4.4.4  requires  php (^7.1.3)
symfony/service-contracts  v2.0.1  requires  php (^7.2.5)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,13 @@ jobs:
       dist: precise
       env: PHPCS_VERSION="2.0.0"
 
+    - php: nightly
+      env: PHPCS_VERSION="dev-master" LINT=1
+
+  allow_failures:
+    # Allow failures for unstable builds.
+    - php: "nightly"
+
   fast_finish: true
 
 before_install:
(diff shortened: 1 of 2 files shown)


Potherca:
Nice work! Merging to master.

PR #109, opened 2020-04-01
Fix case of class name

Seldaek:


jrfnl:
Something weird going on with the Travis status reporting. The original build passed a while back and wasn't reporting back. I restarted the build (still passes as expected) and it's still not reporting back.

Someone with more privileges to this repo than me should have a look at the settings... /cc @Potherca

PR #117, opened 2020-08-19
Update code of conduct

Potherca:
This MR moves the code of conduct from `CONTRIBUTING.md` into a separate file, updates the code-of-conduct to Contributor Covenant v2 and replaces Frencks email with my own for reporting issues.

It also adds a Contributor Covenant badge to README file, to show we adhere to the covenant.


jrfnl (changes requested):
Two remarks:
* For GH to pick up on the file automatically, it should be called `CODE_OF_CONDUCT.md` (underscores, not dashes).
* As we already have a `.github` directory in place, I'd like to suggest moving the file there. That's a fully supported location.

Also see: https://docs.github.com/en/github/building-a-strong-community/adding-a-code-of-conduct-to-your-project#adding-a-code-of-conduct-using-a-template

mjrider (dismissed):

Potherca:
> For GH to pick up on the file automatically, it should be called CODE_OF_CONDUCT.md (underscores, not dashes).

Yeah, that's a derp on my side.

> As we already have a .github directory in place, I'd like to suggest moving the file there. That's a fully supported location.

I personally prefer having the file in the root, making it possible to link to it from markdown (and on Github Pages if/when they are used).

jrfnl (approved):

PR #118, opened 2020-08-19
Document release process

Potherca:
This MR documents the basic steps required to create a new release in the contributing file.

It also add `github_changelog_generator` setting file and a script to run the changelog generator.

Closes #114

PR #122, opened 2020-11-23
Add remark configuration

Potherca:
## Proposed Changes

This MR adds a configuration file for Remark, so there is now one consistent way Markdown should be written in this repo.
The configuration uses the `lint-recommended` preset, with as only exception that the `list-item-indent` has been set to only require a single space after a list character (as opposed to the multi-space default).

Other than a few links in the README, all markdown files in the repo already adhere to this standard.

As there is already a GitHub action that runs Remark, no other changes are needed.

## Related Issues

This MR is inspired by the comment from @jrfnl in #118 regarding inconsistencies in my Markdown.
I suggested adding a Remark config, so here it is! :grin:

### DIFF
--- a/.remarkrc
+++ b/.remarkrc
@@ -1,0 +1,6 @@
+{
+  "plugins": [
+    "remark-preset-lint-recommended",
+    ["remark-lint-list-item-indent", "space"]
+  ]
+}
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ _Note: This plugin is compatible with both version 2.x and 3.x of_ [PHP_CodeSnif
 
 ## Usage
 
-Installation can be done with [composer], by requiring this package as a development dependency:
+Installation can be done with [Composer][composer], by requiring this package as a development dependency:
 
 ```bash
 composer require --dev dealerdirect/phpcodesniffer-composer-installer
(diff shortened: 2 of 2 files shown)


jrfnl (commented):
Good addition.

* Should some information be added to the `CONTRIBUTING` file to document that markdown is expected to conform to the Remark rules ?
* Should some information be added to the `CONTRIBUTING` file on how to locally install and run this check ?
* If the above would suggest installation in a project local manner - are there files which should be added to the `.gitignore` file ?

A quick search yielded me these as relevant links:
* https://github.com/remarkjs/remark-lint
* https://github.com/remarkjs/remark-lint/blob/main/doc/rules.md

Potherca:
Good points. 

I've been thinking about that as well. Not just for Remark, but regarding the other GitHub actions we run... There seems to be a lot of implicit knowledge assumed, which is a bad sign.

We use Pipeline-Components, which are basically just Docker images wrapped around a specific tool.

The options, from top to bottom are:

- Run GitHub actions locally with [act](https://github.com/nektos/act) `act --job lint-remark`
- Run a Pipeline-Component using docker `docker run -v $PWD:/code pipelinecomponents/remark-lint --no-stdout .`
- Run Remark directly with NodeJS

Not sure which route is the best to go and how much to explain or knowledge to assume (i.e. what to explain and where to link to the Remark docs).

I'm inclined to make a separate issue for documenting how to run all of the QA tools in our current pipeline and continue this conversation there...

jrfnl (approved):

jrfnl (dismissed):
Approving under the understanding that "how to run the tooling" will be documented in a separate PR.

Potherca:
Merging without re-review as there are no new changes, just a rebase of already reviewed changes.

PR #124, opened 2020-12-04
Travis: add build against PHP 8.0

jrfnl:
## Proposed Changes

PHP 8.0 has been branched off two months ago, so `nightly` is now PHP 8.1 and in the mean time PHP 8.0 was released last week.

As of today, there is a PHP 8.0 image available on Travis.

This PR adds two new builds against PHP 8.0 to the matrix and, as PHP 8.0 has been released, these builds are not allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,12 @@ jobs:
         - composer validate
 
     - stage: test
-      php: 7.4
+      php: 8.0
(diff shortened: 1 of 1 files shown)


Potherca (approved):
Love it! Merge it like its hot!

Issue #125, opened 2020-12-07
Order of `installed_paths` inconsistent between runs


kevinfodness:
## Problem/Motivation

> I am using this plugin to register several phpcs sniffs from different repos into my `installed_paths` with phpcs. I am also using Travis CI's caching directives in combination with phpcs' caching directives in order to cache the results of scans between runs for files that have not been modified. However, it appears that this plugin doesn't register `installed_paths` in a consistent way between CI runs, which invalidates the phpcs cache, thus negating the benefits of caching between runs.

## Expected behaviour

> Runs of this plugin between CI jobs should produce consistent output for the value of `installed_paths` (specifically, the order of the paths).

## Actual behaviour

> The order of the paths in `installed_paths` is not always the same between CI runs. This is likely due to a race condition of when certain packages finish installing.

## Steps to reproduce

> Use a standard in composer.json that references multiple standards, or reference multiple standards in composer.json (see example below). Create a Travis CI job to run phpcs and echo the value of `installed_paths`. Note that it will not always be in the same order between runs.
>
> Example for run 1:
>
> ```
> "installed_paths": "../../alleyinteractive/alley-coding-standards,../../wp-coding-standards/wpcs,../../sirbrillig/phpcs-variable-analysis,../../automattic/vipwpcs",
> ```
>
> Example for run 2:
>
> ```
> "installed_paths": "../../sirbrillig/phpcs-variable-analysis,../../automattic/vipwpcs,../../alleyinteractive/alley-coding-standards,../../wp-coding-standards/wpcs",
> ```
>
> Contents of composer.json:
>
> ```json
> {
>   "require-dev": {
>     "alleyinteractive/alley-coding-standards": "^0.2.0"
> },
> ```
>
> Contents of composer.json in alley-coding-standards:
>
> ```json
> {
>   "name": "alleyinteractive/alley-coding-standards",
>   "description": "PHPCS sniffs for Alley Interactive",
>   "type": "phpcodesniffer-standard",
>   "license": "GPL-2.0-or-later",
>   "require": {
>     "squizlabs/php_codesniffer": "^3.5.0",
>     "dealerdirect/phpcodesniffer-composer-installer": "*",
>     "wp-coding-standards/wpcs": "^2.3.0",
>     "automattic/vipwpcs": "^2.1.0",
>     "phpcompatibility/phpcompatibility-wp": "*"
>   },
>   "require-dev": {}
> }

## Proposed changes

[...]

PR #127, opened 2020-12-07
Stabilize a condition

jrfnl:
## Proposed Changes

While looking into the (non-)race condition described in #125, I noticed a potentially related issue.

As per the documentation of [realpath()](https://www.php.net/manual/en/function.realpath.php#refsect1-function.realpath-returnvalues):
> realpath() returns FALSE on failure, e.g. if the file does not exist.

In the rare case that `realpath()` _would_ return `false` within this plugin, `false` would subsequently be passed on to `is_dir()` and `is_readable()` which both expect a string path.

This tiny change makes the condition fail earlier by checking the return value of `realpath()`.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -410,7 +410,11 @@ class Plugin implements PluginInterface, EventSubscriberInterface
 
             if (
                 (is_dir($path) === false || is_readable($path) === false) &&
-                (is_dir($alternativePath) === false || is_readable($alternativePath) === false)
+                (
+                    $alternativePath === false ||
+                    is_dir($alternativePath) === false ||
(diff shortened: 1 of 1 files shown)


Potherca (approved):

jrfnl:
@Potherca Just checking: where there any changed you wanted ? Or did you assign this to me to self-merge ?
(sorry, the commit process in this repo still confuses me - I'm used to never self-merge when collaborating, i.e. the (last) approver merges)

Potherca:
No changes needed. I just didn't feel like waiting around for Travis to complete but also didn't want to use my admin right to merge _again_. I assigned you in case I forgot to come back to this :joy:

PR #140, opened 2021-12-23
CI: various updates

jrfnl:
### CS: minor cleanup

PHPCS 3.6.2 added a sniff for a PSR-12 rule which was previously not strictly checked: "No blank line after the opening brace of a class".

This fixes the newly flagged issues.

### GH Actions: auto-cancel previous builds for same branch

Previously, in Travis, when the same branch was pushed again and the "Auto cancellation" option on the "Settings" page had been turned on (as it was for most repos), any still running builds for the same branch would be stopped in favour of starting the build for the newly pushed version of the branch.

To enable this behaviour in GH Actions, a `concurrency` configuration needs to be added to each workflow for which this should applied to.

More than anything, this is a way to be kind to GitHub by not wasting resources which they so kindly provide to us for free.

Refs:
* https://github.blog/changelog/2021-04-19-github-actions-limit-workflow-run-or-job-concurrency/
* https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

### GH Actions: use error_reporting=-1

... as `E_ALL` does not always contain _all_ errors across PHP versions.

### GH Actions: update the security checker

Ref: https://github.com/fabpot/local-php-security-checker/blob/main/CHANGELOG.md

### 🆕 GH Actions: version update for `ramsey/composer-install`

The action used to install Composer packages and handle the caching has released a new major (and some follow-up patch releases), which means, the action reference needs to be updated to benefit from it.

Refs:
* https://github.com/ramsey/composer-install/releases/tag/2.0.0
* https://github.com/ramsey/composer-install/releases/tag/2.0.1
* https://github.com/ramsey/composer-install/releases/tag/2.0.2

### 🆕 GH Actions: always use --no-interaction for Composer

Adding `--no-interaction` to "plain" Composer commands to potentially prevent CI hanging if, for whatever reason, interaction would be needed in the future.

### 🆕 🆕 GH Actions: version update for `actions/checkout`

Ref: https://github.com/actions/checkout/releases

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -34,7 +34,6 @@ use Symfony\Component\Process\PhpExecutableFinder;
  */
 class Plugin implements PluginInterface, EventSubscriberInterface
 {
-
     const KEY_MAX_DEPTH = 'phpcodesniffer-search-depth';
 
     const MESSAGE_ERROR_WRONG_MAX_DEPTH =
--- a/.github/workflows/integrationtest.yml
+++ b/.github/workflows/integrationtest.yml
@@ -151,7 +151,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          ini-values: error_reporting=E_ALL, display_errors=On
+          ini-values: error_reporting=-1, display_errors=On
           coverage: none
 
       - name: 'Composer: set PHPCS version for tests'
--- a/.github/workflows/integrationtest.yml
+++ b/.github/workflows/integrationtest.yml
@@ -155,12 +155,12 @@ jobs:
           coverage: none
 
       - name: 'Composer: set PHPCS version for tests'
-        run: composer require --no-update --no-scripts squizlabs/php_codesniffer:"${{ matrix.phpcs_version }}"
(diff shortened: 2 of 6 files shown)


src/Plugin.php
```diff
@@ -34,7 +34,6 @@
  */
 class Plugin implements PluginInterface, EventSubscriberInterface
 {
-
```
Potherca, line 37: [That's a fair cop!](https://en.m.wiktionary.org/wiki/fair_cop) :joy:

Potherca (approved):

jrfnl:
Added two extra commits for your review (and updated the PR description).

Potherca (approved):

Potherca:
Looks good.

jrfnl:
Found yet another snippet which needed updating. Added a new commit with the additional change.

Potherca (approved):

jrfnl:
Anything I can do to move this PR forward ?

Potherca:
I was suffering from a severe lack of time, which has now been resolved. Will look into this the coming few days :+1:

PR #141, opened 2021-12-23
README: update with information about Composer >= 2.2

jrfnl:
### README: update with information about Composer >= 2.2

As of Composer 2.2.0, Composer plugins need to be explicitly allowed to run.

This updates the readme to inform users of this mechanism.

Ref: https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution

### README: update coding standards dev code sample

... to be a little more realistic for current dev.

### 🆕 README: add information about adding permission via CLI

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,11 @@ When permission has been granted, the following snippet will automatically be ad
 }
 ```
 
+When using Composer < 2.2, you can add the permission flag ahead of the upgrade to Composer 2.2, by running:
+```bash
+composer config allow-plugins.dealerdirect/phpcodesniffer-composer-installer true
+```
+
 That's it.
 
 ### Compatibility
--- a/README.md
+++ b/README.md
@@ -25,6 +25,19 @@ Installation can be done with [Composer][composer], by requiring this package as
 composer require --dev dealerdirect/phpcodesniffer-composer-installer
 ```
 
+When using Composer 2.2 or higher, Composer will [ask for your permission](https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution) to allow this plugin to execute code. For this plugin to be functional, permission needs to be granted.
+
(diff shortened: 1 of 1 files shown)


Potherca (dismissed):

Potherca:
I was wondering if, based on this, _we_ need to add `--no-interaction` anywhere in our pipelines? :thinking:

jrfnl:
> I was wondering if, based on this, _we_ need to add `--no-interaction` anywhere in our pipelines? 🤔

We don't (need to).

The thing is, when doing a `composer install/update` for _this_ project itself, you don't get the question as this _is_ a Composer plugin. The code of this plugin won't run on itself.
I tested as I wondered how this would work for projects which _are_ Composer plugins, but the plugin is not run on itself and behaviour is the same as before.

That's also why in the test scrips in the CI, there a separate call to `composer install-codestandards` to run the script.

Other than that, we use `ramsey/composer-install` in most places and that will always add `--no-interaction`to the command.

jrfnl:
Reviewed the scripts now and found a few places we _could_ add `--no-interaction`. Commit added to PR #140.

Potherca:
Good work!

jrfnl:
Inspired by a similar README update I saw in the [Composer Normalize README](https://github.com/ergebnis/composer-normalize#installation), I've added one more commit which shows the CLI command to add the permission.

Potherca (approved):

jrfnl:
Anything I can do to move this PR forward ?

Potherca:
It was mostly a lack of time that kept this standing still.

I had a deadline at work that kept me busy throughout the holidays, and I had to help the kids with school as we were in quarantine. This coming week the kids are back to school and I have time of from work (we made the deadline :tada:), so I plan on catching up here and in other projects.

PR #147, opened 2022-01-26
Drop support for PHP 5.3

### LINKED ISSUE #145: Drop support for 5.3
Based on the usage statistics from Packagist and supported versions from PHP_CS we no longer need to support 5.3.

Support can be dropped before we label a stable 1.0 version

jrfnl:
## Proposed Changes

This PR just and only drops support for PHP 5.3. It doesn't (yet) take advantage of features which were introduced in PHP 5.4, which could now be used.

## Related Issues

Fixes #145

### DIFF
--- a/.github/workflows/integrationtest.yml
+++ b/.github/workflows/integrationtest.yml
@@ -29,7 +29,7 @@ jobs:
 
         include:
           # Ensure a "highest" PHP/PHPCS build combination for PHPCS 2.x is included.
-          - php: '5.3'
+          - php: '5.4'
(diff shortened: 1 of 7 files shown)


jrfnl:
Note: when a decision is taken on this, the required checks for the branch (branch protection settings) will need updating as well.

Potherca (approved):

Potherca (approved):

jrfnl:
FYI: I've updated the branch protection settings for the changes in this PR now.

PR #152, opened 2022-01-30
Composer: tweak PHPCS version constraint

jrfnl:
## Proposed Changes

Some of the early PHPCS 3.x versions did not support external standards correctly, so using this plugin with those would always run into trouble, so let's just make it explicit that `3.0.0` and `3.0.1` are not supported.

`3.0.2` should be fine, though IIRC, even `3.1.0` contained some fixes for external standard support, so we could choose to up this to `^3.1.0` to be on the safe side.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
   "require": {
     "php": ">=5.3",
     "composer-plugin-api": "^1.0 || ^2.0",
-    "squizlabs/php_codesniffer": "^2.0 || ^3.0 || ^4.0"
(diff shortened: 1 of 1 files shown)


Potherca (approved):

jrfnl:
Just came across issue #33 and based on the discussion in that issue, which references upstream fixes needing to be made which only got released in PHPCS 31.0, I do think upping the minimum for the 3.x branch to `3.1.0` seems prudent. I've updated the PR to reflect this.

Potherca (approved):

PR #157, opened 2022-02-09
README: remove references to Scrutinizer

jrfnl:
## Proposed Changes

Looks like the connection with Scrutinizer was broken more than 3 years ago, so the score is by now out of date and because of that irrelevant.

Ref: https://scrutinizer-ci.com/g/dealerdirect/phpcodesniffer-composer-installer/

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,6 @@
 [![License][license-shield]](LICENSE.md)
 
 [![Tests][ghactionstest-shield]][ghactions]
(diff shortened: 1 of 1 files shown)


Potherca (approved):

jrfnl:
Note: Build failure against PHP 8.2 is unrelated to this PR (actually a problem with `setup-PHP`) and shouldn't be a blocker for merging this PR.

PR #126, opened 2020-12-07
Sort list of installed paths before saving for consistency

kevinfodness:
## Proposed Changes

> Fixes the inconsistency in `installed_paths` documented in #125 by sorting the list of installed paths alphabetically before saving them.

## Related Issues

> #125

Issue #92, opened 2019-12-17
Iterate further on the integration tests via Travis


jrfnl:
From #91:

> Future improvements
> 
> This is a first iteration to improve the build script.
> 
> In a further iteration, I'd like to suggest exploring the following additional variants:
> 
> * Having some builds with a global install of PHPCS + global install of the external standard used for testing.
> * Having some builds with a mix of global/local installs of external standards(s).
> * Having some builds using Windows as OS instead of Linux.

Potherca:
A couple of years ago I create [some code for phpcodesniffer-composer-installer issue-24](https://github.com/potherca-abandoned/phpcodesniffer-composer-installer_issue-24) that might be usable for some of this.

Potherca:
Not sure how/where to best run build steps in Windows. Maybe @mjrider has some ideas?

jrfnl:
@Potherca I know you can set the OS for specific builds in both Travis as well as actions, so either could work.

jrfnl:
Especially with an eye on preventing regression, while still broadening support to newer versions, like PHPCS 4.x, PHP 8.x and Composer 2.x, I've made an inventory of test cases which would need to be included by going through _every single_ issue and PR in this repo from the very beginning to the time of writing.

My focus has been on integration tests and I have made "shorthand" notes.
I don't have time at this moment to get this setup completely, nor have I a clear idea on how we can set up these type of CLI integration tests easily.

All the same, I figured it may be useful to make these notes public for whomever starts work on getting tests set up when they have time.

----

OS matrix
- Windows
- Unix

Travis matrix:
- PHPCS 2 low/high - PHP 5.3 - PHP 8
- PHPCS 3 low/high - PHP 5.4 - PHP 8
- PHPCS 4 (allowed to fail) - PHP 7.2 - PHP 8
- Composer 1
- Composer 2

Possibly use:
--prefer-lowest (works with update/require)
--prefer-stable (works with update/require)
--prefer-source (works with update/require/install)



Basic test scenarios:
- Local install
- Global install

More detailed test scenarios:
- With another dependency which has a file autoload (Symfony polyfill)
- In combination with other Composer plugins
- In combination with post-install/update-cmd's (see #105)

Note: When error output is expected, test with both `--verbose` and without and verify the output.


Caveats:
As the tests would include tests which change the global composer install, running tests locally should be done in a VM/Docker container.

[...]

PR #176, opened 2022-04-18
Tests: add new `InstalledPathsOrderTest`

jrfnl:
## Proposed Changes

This new test class tests that the plugin always registers the `installed_paths` in the same order.

I have confirmed that the tests would fail on various CI builds without the fix from #126.

Notes:
* This test is about Composer and the plugin, so does not need to be tested against multiple PHPCS versions.

This test class covers the following bug previously reported:
* Dealerdirect/phpcodesniffer-composer-installer#125
* Dealerdirect/phpcodesniffer-composer-installer#126

## Related Issues

Related to #92

### DIFF
--- a/tests/IntegrationTest/InstalledPathsOrderTest.php
+++ b/tests/IntegrationTest/InstalledPathsOrderTest.php
@@ -1,0 +1,127 @@
+<?php
+
+/**
+ * This file is part of the Dealerdirect PHP_CodeSniffer Standards
+ * Composer Installer Plugin package.
+ *
+ * @copyright 2022 PHPCodeSniffer Composer Installer Contributors
+ * @license MIT
+ */
+
+namespace Dealerdirect\Composer\Plugin\Installers\PHPCodeSniffer\Tests\IntegrationTest;
+
+use Dealerdirect\Composer\Plugin\Installers\PHPCodeSniffer\Plugin;
+use Dealerdirect\Composer\Plugin\Installers\PHPCodeSniffer\Tests\TestCase;
+
+/**
+ * Test that the plugin always registers the installed_paths in the same order.
+ *
+ * This test is about Composer and the plugin, so does not need to be tested against multiple PHPCS versions.
+ *
+ * @link https://github.com/PHPCSStandards/composer-installer/issues/125
+ * @link https://github.com/PHPCSStandards/composer-installer/pull/126
+ */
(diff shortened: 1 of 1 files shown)


tests/IntegrationTest/InstalledPathsOrderTest.php
```diff
@@ -0,0 +1,119 @@
...
+         * 3. Retrieve the installed paths from both and compare to ensure the order is the same.
+         */
+        $globalPaths = $this->executeCliCommand('"vendor/bin/phpcs" --config-show', static::$tempGlobalPath);
+        $this->assertSame(0, $globalPaths['exitcode'], 'Exitcode for "phpcs --config-show" did not match 0 (global)');
+
+        $localPaths = $this->executeCliCommand('"vendor/bin/phpcs" --config-show', static::$tempLocalPath);
+        $this->assertSame(0, $localPaths['exitcode'], 'Exitcode for "phpcs --config-show" did not match 0 (local)');
+
+        // Get the installed paths setting from the config.
+        $this->assertSame(1, preg_match('`installed_paths:\s+([^\n\r]+)\s+`', $globalPaths['stdout'], $matchGlobal));
```
Potherca, line 109: Something irks me about the `assertSame` / `preg_match` combination. Not entirely sure what... Could `assertMatchesRegularExpression` be used? :thinking:

Potherca (dismissed):

tests/IntegrationTest/InstalledPathsOrderTest.php
jrfnl: Good point. I think I may have created this test before we decided to drop PHP 5.3 support and add the PHPUnit Polyfills. Will fix this up.

jrfnl: Actually, looking at it now, I _do_ see why I didn't use `assertMatchesRegularExpression()`: we need to remember the `$matches` for handling and use in the _next_ assertion and `assertMatchesRegularExpression()` doesn't have a way to give us access to that.

jrfnl: Rebased & force pushed the fixed commit. The additional `$message` parameters in the last three assertions is the only change.

Potherca (approved):

jrfnl:
Rebased without changes after the merge of #180 to get a clean and passing build. Will merge after the build passes.

PR #184, opened 2022-06-24
GH Actions: run linting against against ubuntu-latest

jrfnl:


Potherca:
For the sake of future sanity, I'd rather pin this to `ubuntu-20.04`.  Support for `ubuntu-22.04` [is still in beta](https://github.com/actions/virtual-environments) but whenever it is moved to `latest` by GitHub, I'd rahter not see things accidentally break. Or we could move to 22.4 now and not have to worry about updating things for at least another four years 🤷 

Either way, I'd rather not use `latest`

jrfnl:
@Potherca We're using `latest` for the tests and to be fair, the ubuntu version shouldn't really matter much for the checks being done here. Happy enough to pin it to `20-04` though.

Potherca (approved):

Potherca:
No, you are right... `latest` is fine here. The linter is not run on the host but in a container on the host.

I got my wires crossed somehow... :facepalm:

PR #189, opened 2022-10-03
Correct grammar in error message

fredden:
## Proposed Changes

The error message shown when an invalid value is provided for `extra.phpcodesniffer-search-depth` in `composer.json` had incorrect grammar. This pull request fixes this.

https://www.merriam-webster.com/words-at-play/when-to-use-then-and-than


jrfnl (approved):
Thank you for contributing! Change looks good to me.

PR #191, opened 2022-10-03
Rename namespace prefix from `Dealerdirect`  to `PHPCSStandards`

### LINKED ISSUE #188: Rename the namespace
Previously mentioned in comments in #159

The package has moved to a new organisation - see #146.

For consistency and predictability as well as not to confuse people less familiar with the history of the project, I'd like to propose updating the namespace.

I'm undecided on what the new namespace should be.

The other projects in the `PHPCSStandards` organisation do not use the `PHPCSStandards` namespace prefix, they mostly use a project name based prefix, though this is largely due to the namespace rules from PHPCS having to be respected for those projects.

**Open question**:
* Are there any namespace name rules for Composer plugins which should be followed ?

If not, just swopping out the `DealerDirect` prefix in favour of `PHPCSStandards` seems the most straight-forward way to action this.

jrfnl:
## Proposed Changes

As discussed in issue #188.

**Important**: this needs a "breaking change" changelog entry as the command to run the plugin directly (either manually or via a script) has changed. See the README updates for the change which needs to be highlighted in the changelog.

## Related Issues

Fixes #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ section of the `composer.json`:
 {
     "scripts": {
         "install-codestandards": [
-            "Dealerdirect\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run"
+            "PHPCSStandards\\Composer\\Plugin\\Installers\\PHPCodeSniffer\\Plugin::run"
         ]
     }
 }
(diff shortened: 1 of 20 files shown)


fredden:
Is changing the package name in `composer.json` in-scope for this, or will that be handled separately?
https://github.com/PHPCSStandards/composer-installer/blob/306335a12f6cfa98cdd0681d874f9eacfe7b652d/composer.json#L2

Potherca (approved):
Also checked locally, could not find anything missing. :+1:

PR #194, opened 2022-12-11
GH Actions: selectively use `fail-fast` with setup-php

jrfnl:
## Proposed Changes

I've seen some recent build failures (elsewhere) due to the `setup-php` action running into a rate limit and not downloading the required version of Composer. In the case of this package, that would make the test runs worthless.

The `setup-php` action runner defaults to _showing_ these type errors in the logs, but not stopping the workflow run.

So, specifically for those jobs where the Composer version is important, I'm adding the `fail-fast` option to `setup-php` to fail the build if the action runner ran into any errors.

Ref: https://github.com/shivammathur/setup-php#fail-fast-optional


Potherca (approved):
👍

PR #198, opened 2023-01-02
.gitattributes: sync with current repo state

jrfnl:
## Proposed Changes

A number of dev-related files/directories were not listed in the `.gitattributes` file, while these don't need to be included in the distribution archives.

Fixed now.


Potherca (approved):

Potherca:
Merged without waiting for all checks passing as only non-code changes were made.

PR #199, opened 2023-01-02
Rename references to `master` branch

### LINKED ISSUE #160: Open discussion: rename the default branch ?
In a way a "bit late to the party", but should the default branch be renamed away from `master` ?

And if so, what would be the preferred new name ?

Some options:
* A lot of repos have gone with `main`. For a "one branch" repo, that would be an obvious choice.
* As an alternative, but that would also involve a different branching strategy, it could be considered to have a `stable` branch as a release branch and have a `develop` branch as the default PR branch. For a new release `develop` would then be merged into `stable`.

Opinions ?

jrfnl:
## Proposed Changes

This PR is to be merged in conjunction with the branch rename to `main` as discussed in #160

## Related Issues

Fixes #160

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -49,5 +49,5 @@ Output of `vendor/bin/phpcs --config-show`:
 (Please paste the output you get here)
 ```
 
-## Tested against `master` branch?
-- [ ] I have verified the issue still exists in the `master` branch.
+## Tested against `main` branch?
+- [ ] I have verified the issue still exists in the `main` branch.
--- a/.github/workflows/integrationtest.yml
+++ b/.github/workflows/integrationtest.yml
@@ -2,10 +2,10 @@
 name: Integration Test
 
 on:
-  # Run on pushes to `master` and on all pull requests.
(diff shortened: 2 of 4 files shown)


Potherca (approved):
I've checked the code, I didn't find other instances that the ones changed here.

The only other reference is https://github.com/PHPCSStandards/composer-installer/blob/master/tests/PHPCSVersions.php#L23 but if I understand correctly that is the branch on PHPCS not this project.

jrfnl:
> I've checked the code, I didn't find other instances that the ones changed here.
> 
> The only other reference is https://github.com/PHPCSStandards/composer-installer/blob/master/tests/PHPCSVersions.php#L23 but if I understand correctly that is the branch on PHPCS not this project.

100% correct. That's the PHPCS branch, not a reference to a branch in this project. Only thing we could do there, would maybe be to update line 19 to make it crystal clear that is a PHPCS branch reference:
https://github.com/PHPCSStandards/composer-installer/blob/fcae46579e2f44e713f18d3643b0322387e2a3b2/tests/PHPCSVersions.php#L19

Potherca:
I have renamed branch `master` to `main`. These changes have been merged into `main` via https://github.com/PHPCSStandards/composer-installer/pull/201

PR #203, opened 2023-02-09
Change project stage in README to "production read".

Potherca:
This MR changes the badge for "Project Stage" to "Production Ready" (as the project has reached a v1).
A link to the page which explains the rational behind the badge is also added.

**Question to be aswered**: Should we _update_ the badge or does it make more sense to _remove_ it?
Orignally it was added to make clear (in the "swag" section) that the project was not yet production ready, as not everyone seems to understand what `v0` versions seem to mean. However, as the default assumption seems to be that any project _is_ production ready (unless otherwise stated) is the badge still needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # PHP_CodeSniffer Standards Composer Installer Plugin
 
-![Project Stage][project-stage-shield]
+[![Project stage: Production Ready][project-stage-shield]][project-stage-page]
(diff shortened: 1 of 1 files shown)


jrfnl:
@Potherca IMO the use of semver takes care of this, so the badge is not needed and can be removed.
If people don't understand semver, they should educate themselves.

Potherca:
Closed in favor of #204

PR #206, opened 2023-08-18
Replace http with https

SplotyCode:
Proposed Changes

Related Issues
Code scans mark it as insecure. Here it really does not matter as the side upgrades to https it is still good practise

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,9 +12,9 @@
   "authors": [
     {
       "name": "Franck Nijhof",
-      "email": "franck.nijhof@dealerdirect.com",
-      "homepage": "http://www.frenck.nl",
-      "role": "Developer / IT Manager"
+      "email": "opensource@frenck.dev",
+      "homepage": "https://frenck.dev",
+      "role": "Open source developer"
     },
     {
       "name" : "Contributors",
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,6 @@
     "standard", "standards", "style guide", "stylecheck",
     "qa", "quality", "code quality", "tests"
   ],
-  "homepage": "https://www.dealerdirect.com",
   "license": "MIT",
   "authors": [
     {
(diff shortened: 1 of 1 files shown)


jrfnl:
@Potherca As the plugin is no longer directly associated with DealerDirect, I'm thinking we may as well remove the `homepage` key from the `composer.json` file. What do you think ?

Potherca:
I'd say either point it to the repo `https://github.com/PHPCSStandards/composer-installer` or remove it.
Either one is fine by me.

jrfnl:
@Potherca As the `composer.json` file already contains a link to the repo in `support` -> `source`, I don't think `homepage` linking to the repo has any added benefit.

@SplotyCode Thanks for the PR, would you mind removing the `homepage` key instead ?

composer.json
SplotyCode, line 1: [www.frenck.nl ](http://www.frenck.nl) was offline when i opened the pr and is still offline

jrfnl: Hmm... I've send @frenck a message about that. If its doesn't come up again, we may want to remove that link too.

frenck: It has moved to `https://frenck.dev`

frenck: It is pretty much outdated as a whole; I'm a happy full-time open source developer nowadays :)

If you want, feel free to update it to: 

```json
    {
      "name": "Franck Nijhof",
      "email": "opensource@frenck.dev",
      "homepage": "https://frenck.dev",
      "role": "Open source developer"
    },
```

I'm happy to send in a PR if that is wanted instead.

../Frenck

frenck: PS: I feel like there are quite a few key authors missing in this file... 🙈

jrfnl (approved):
Thanks for making those updates @SplotyCode ! All looks good to me now.

jrfnl:
@Potherca Could you please have another look & merge if you're also okay with the changes as they are now ?

Potherca (approved):

PR #213, opened 2023-11-26
GH Actions/quicktest: fix the build

jrfnl:
## Proposed Changes

Builds against PHP 5.4 on Windows are failing. The failure happens even before the test run at the "Composer install" step.

These failures look to be unrelated to this package, but seem to be related to PHP 5.4 trying to make an unsecured connection to Packagist and this no longer being supported. There may be ways to work around this, but I haven't been able to find one other than downloading a new certificate bundle and using that (not tested).

Considering PHP 5.4 is ancient, I'm proposing to just drop the test runs against PHP 5.4 in combination with Windows.


jrfnl:
To merge this PR, the branch protection needs to be updated to remove the two PHP 5.4/Windows builds from being required.

PR #218, opened 2024-04-19
Composer: avoid writing a lock file

fredden:
## Proposed Changes

When working with this repository as a developer, we should be using the latest compatible packages. By writing a lock file for Composer, we may get into a state where we are "stuck" on an older version of a dependency. This change avoids such a situation by telling Composer to not write out a lock file in the project.

### DIFF
--- a/.github/workflows/securitycheck.yml
+++ b/.github/workflows/securitycheck.yml
@@ -32,6 +32,9 @@ jobs:
           php-version: ${{ matrix.php }}
           coverage: none
 
+      - name: Enable creation of `composer.lock` file
+        run: composer config --unset lock
+
       # Install dependencies and handle caching in one go.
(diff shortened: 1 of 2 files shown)


jrfnl (changes requested):
Thanks @fredden for this PR! I agree this is a good change.

To allow for merging this PR, however, a small tweak need to be made to the `securitycheck.yml` workflow which relies on the `composer.lock` file.
 I think an extra step needs to be added to the workflow, before the "compose install" step, to remove this key from the config. that should allow for the workflow to continue to work and be useful.

jrfnl (approved):
LGTM 👍🏻

Potherca (approved):

jrfnl:
Merging. The build failures against PHP 8.4 are unrelated to this PR and will be addressed separately via PR #219.

PR #225, opened 2024-08-22
GH Actions: show deprecations when linting & lint against next PHP version

jrfnl:
### GH Actions: show deprecations when linting

While rare, there are some deprecations which PHP can show when a file is being linted.
By default these are ignored by PHP-Parallel-Lint.

Apparently though, there is an option to show them (wasn't documented until recently), so let's turn that option on.

### GH Actions: also lint against _next_ PHP version

The PHP lint job currently checks all PHP files for parse/compile errors against a limited set of PHP versions, including the high/low versions.

It did not run the linter against the _next_ (upcoming) PHP version yet.

This has now been added with the `'nightly'` alias, which means that the version will automatically roll through to the next PHP version when a new PHP branchs gets cut, without us needing to manually maintain this.

Also note that the build against `nightly` is allowed to fail.

### DIFF
--- a/.github/workflows/phplint.yml
+++ b/.github/workflows/phplint.yml
@@ -19,7 +19,9 @@ jobs:
 
     strategy:
       matrix:
-        php: ['5.4', '5.6', '7.2', 'latest']
+        php: ['5.4', '5.6', '7.2', 'latest', 'nightly']
+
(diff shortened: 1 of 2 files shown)


Potherca (approved):
👍

Issue #221, opened 2024-05-03
Proposal for limiting supported versions


jrfnl:
At this time, the plugin is compatible with:
- PHP **5.4+**, **7.x**, and **8.x** (Support for PHP v8 is available since `v0.7.0`)
- [Composer][composer] **1.x** and **2.x** (Support for Composer v2 is available since `v0.7.0`)
- [PHP_CodeSniffer][codesniffer] **2.x** and **3.x** (Support for PHP_CodeSniffer v3 is available since `v0.4.0`)

These are very wide ranges and is a compatibility span of more than a decade for most of these.

With this ticket, I'd like to open the discussion for limiting the range of supported versions a little.

### Proposal

I would like to propose to drop support for:
* PHP < 7.2
* Composer 1.x
* PHP_CodeSniffer 2.x

### Planning

I propose that these changes be made once PHP_CodeSniffer 4.0 has been released (which also drops support for PHP < 7.2).

I also propose for these changes to be released as version 1.1.0 of the plugin.

### Consequences

As users of this plugin will, by definition, be using Composer, we would not be leaving users behind as Composer will handle the version negotiation and can still install the 1.0.x version if necessary.

Based on the [PHP version stats available via Packagist](https://packagist.org/packages/dealerdirect/phpcodesniffer-composer-installer/php-stats), the PHP version drop would affect 1.5% of the users of this plugin.

While not necessarily a precise representation of the users of this plugin, based on the [PHP_CodeSniffer version stats available via Packagist](https://packagist.org/packages/squizlabs/php_codesniffer/stats), the PHP_CodeSniffer version drop would affect less than 0.2% of all potential users of this plugin.

Opinions ?

fredden:
This sounds like a sensible change to me. For similar reasons (and considering the stability of this package), I would also be comfortable with adjusting the minimum supported PHP to 8.2 (https://www.php.net/supported-versions), and Composer to 2.6. Users who still require a version which works on older systems can use already-released versions of this package, and Composer is well placed to resolve this automatically.

Releasing this change as a minor version makes sense given the explanations above.

jrfnl:
In my opinion, upping the minimum supported PHP version to anything beyond PHP 7.2 is not up for discussion.

This plugin hooks into PHP_CodeSniffer and is used by lots of external standards for PHP_CodeSniffer.

[...]

PR #229, opened 2024-11-27
Ignore development-only packages in security check

fredden:
## Proposed Changes

Only perform security check on actual dependencies / ignore development-only dependencies for this check.

## Related Issues

https://github.com/PHPCSStandards/composer-installer/pull/228#issuecomment-2492542642

### DIFF
--- a/.github/workflows/securitycheck.yml
+++ b/.github/workflows/securitycheck.yml
@@ -35,9 +35,6 @@ jobs:
           php-version: ${{ matrix.php }}
           coverage: none
 
-      - name: Enable creation of `composer.lock` file
-        run: composer config --unset lock
-
       # Install dependencies and handle caching in one go.
       # @link https://github.com/marketplace/actions/install-php-dependencies-with-composer
       - name: Install Composer dependencies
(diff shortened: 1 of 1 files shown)


jrfnl (dismissed):
Makes sense to me. Thanks @fredden!

jrfnl:
🤔 Hmm... based on the build on your fork, this doesn't solve it for PHP 5.4...? Composer still creates the `lock` file, even if the packages aren't being installed and the `lock` file is what's used by the Security check...

fredden:
> 🤔 Hmm... based on the build on your fork, this doesn't solve it for PHP 5.4...? Composer still creates the `lock` file, even if the packages aren't being installed and the `lock` file is what's used by the Security check...

I've updated the check command to use the actually-installed packages instead.

jrfnl (approved):
Interesting. Looks like checking the `installed.json` file is an [undocumented option](https://github.com/fabpot/local-php-security-checker).

Also noticed that the Local security checker package is now deprecated in favour of the `composer audit` command.

We should probably update this workflow to use that once support for PHP < 7.2 has been dropped (issue #221).

fredden:
It turns out that this checker tool also has a `--no-dev` flag which could have been used here. If you would prefer that method, let me know and I will update this pull request accordingly.

fredden:
The failing tests here will be solved by https://github.com/PHPCSStandards/composer-installer/issues/221

jrfnl:
@Potherca Do you still want to have a look at this or can I merge it ?

Potherca:
Merge at will.

PR #230, opened 2025-01-13
Drop support for Composer v1.x

fredden:
## Proposed Changes

Composer themselves will no longer support version 1.x beyond 1st August 2025. Composer version 2.0 was released in October 2020, which is more than four years ago. Anyone wanting to use this plug-in with Composer v1.x can use any existing version.

https://blog.packagist.com/composer-2-0-is-now-available/
https://blog.packagist.com/deprecating-composer-1-support/
https://blog.packagist.com/shutting-down-packagist-org-support-for-composer-1-x/

## Related Issues

- This is step one of three for https://github.com/PHPCSStandards/composer-installer/issues/221

### DIFF
--- a/.github/workflows/quicktest.yml
+++ b/.github/workflows/quicktest.yml
@@ -32,14 +32,15 @@ jobs:
           - '7.2'
           - 'latest'
         composer:
-          - 'v1'
+          # Note: for PHP 5.4 - 7.1, "v2" will install Composer 2.2.
+          # For PHP 7.2+, it will install Composer "latest".
           - 'v2'
         os:
           - 'ubuntu-latest'
           - 'windows-latest'
 
         # Installing on Windows with PHP 5.4 runs into all sorts of problems (which are not ours).
-        # Exclude the Windows PHP 5.4 builds and replace them with PHP 5.5/5.6 for the same.
+        # Exclude the Windows PHP 5.4 builds and replace them with PHP 5.5 for the same.
         # @link https://github.com/PHPCSStandards/composer-installer/issues/181
         # @link https://github.com/PHPCSStandards/composer-installer/pull/213
         exclude:
@@ -47,10 +48,6 @@ jobs:
             os: 'windows-latest'
         include:
           - php: '5.5'
-            composer: 'v2'
-            os: 'windows-latest'
-          - php: '5.6'
-            composer: 'v1'
             os: 'windows-latest'
 
     name: "Quick test"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,12 +32,12 @@ Even better: You could submit a pull request with a fix / new feature!
 ### (Code) Quality checks
 
 Every merge-request triggers a build process which runs various checks to help
-maintain a quality standard. All JSON, Markdown, PHP, and Yaml files are 
+maintain a quality standard. All JSON, Markdown, PHP, and Yaml files are
 expected to adhere to these quality standards.
 
 These tools fall into two categories: PHP and non-PHP.
 
-### PHP 
+### PHP
 
 The PHP specific tools used by this build are:
 
@@ -99,7 +99,7 @@ In case the test setup has trouble locating your `composer.phar` file:
     </php>
     ```
     **Note**: this setting also allows for locally testing with different versions of Composer.
-    You could, for instance, have multiple Composer Phar files locally, `composer1.phar`, `composer2.1.phar`, `composer2.2.phar`.
+    You could, for instance, have multiple Composer Phar files locally, `composer.phar`, `composer2.2.phar`, `composer2.6.phar`.
     By changing the path in the value of this `env` setting, you can switch which version will be used in the tests.
 
 ### Non-PHP
@@ -112,7 +112,7 @@ The non-PHP specific tools used by this build are:
 
 These tools are also run as [GitHub actions][].
 All the checks can be run locally using [`act`][].
-Alternatively they can be run using `docker run`, as all checks use Docker 
+Alternatively they can be run using `docker run`, as all checks use Docker
 images provided by [Pipeline-Component][].
 
 Finally, they could be run locally using NodeJS, Ruby, PHP, or whatever the tool
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Installation can be done with [Composer][composer], by requiring this package as
 composer require --dev dealerdirect/phpcodesniffer-composer-installer:"^1.0"
 ```
 
-When using Composer 2.2 or higher, Composer will [ask for your permission](https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution) to allow this plugin to execute code. For this plugin to be functional, permission needs to be granted.
+Since Composer 2.2, Composer will [ask for your permission](https://blog.packagist.com/composer-2-2/#more-secure-plugin-execution) to allow this plugin to execute code. For this plugin to be functional, permission needs to be granted.
 
 When permission has been granted, the following snippet will automatically be added to your `composer.json` file by Composer:
 ```json
@@ -36,7 +36,7 @@ When permission has been granted, the following snippet will automatically be ad
 }
 ```
 
-When using Composer < 2.2, you can add the permission flag ahead of the upgrade to Composer 2.2, by running:
+You can safely add the permission flag (to avoid Composer needing to ask), by running:
 ```bash
 composer config allow-plugins.dealerdirect/phpcodesniffer-composer-installer true
 ```
@@ -48,7 +48,7 @@ That's it.
 This plugin is compatible with:
 
 - PHP **5.4+**, **7.x**, and **8.x** (Support for PHP v8 is available since [`v0.7.0`][v0.7])
-- [Composer][composer] **1.x** and **2.x** (Support for Composer v2 is available since [`v0.7.0`][v0.7])
+- [Composer][composer] **2.2+** (Support for Composer v2 is available since [`v0.7.0`][v0.7]; support for Composer < 2.2 was dropped in [`v1.1.0`][v1.1])
 - [PHP_CodeSniffer][codesniffer] **2.x** and **3.x** (Support for PHP_CodeSniffer v3 is available since [`v0.4.0`][v0.4])
 
 ### How it works
@@ -283,3 +283,4 @@ THE SOFTWARE.
 [using-composer-plugins]: https://getcomposer.org/doc/articles/plugins.md#using-plugins
 [v0.4]: https://github.com/PHPCSStandards/composer-installer/releases/tag/v0.4.0
 [v0.7]: https://github.com/PHPCSStandards/composer-installer/releases/tag/v0.7.0
+[v1.1]: https://github.com/PHPCSStandards/composer-installer/releases/tag/v1.1.0
--- a/composer.json
+++ b/composer.json
@@ -28,13 +28,13 @@
   },
   "require": {
     "php": ">=5.4",
-    "composer-plugin-api": "^1.0 || ^2.0",
+    "composer-plugin-api": "^2.2",
     "squizlabs/php_codesniffer": "^2.0 || ^3.1.0 || ^4.0"
   },
   "require-dev": {
     "ext-json": "*",
     "ext-zip": "*",
-    "composer/composer": "*",
+    "composer/composer": "^2.2",
     "phpcompatibility/php-compatibility": "^9.0",
     "php-parallel-lint/php-parallel-lint": "^1.3.1",
     "yoast/phpunit-polyfills": "^1.0"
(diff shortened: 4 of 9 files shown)


.github/workflows/quicktest.yml
```diff
@@ -32,25 +32,20 @@ jobs:
           - '7.2'
           - 'latest'
         composer:
-          - 'v1'
           - 'v2'
```
jrfnl, line 35: Minor suggestion - might be helpful to add a small comment here along the lines of:
```
# Note: for PHP 5.4 - 7.1, "v2" will install Composer 2.2. For PHP 7.2, it will install Composer "latest".
```

composer.json
```diff
@@ -28,7 +28,7 @@
   },
```
jrfnl, line 28: Missing: update to the version constraint for the `composer/composer` package in  `require-dev`.

README.md
```diff
@@ -48,7 +48,7 @@ That's it.
 This plugin is compatible with:
 
 - PHP **5.4+**, **7.x**, and **8.x** (Support for PHP v8 is available since [`v0.7.0`][v0.7])
-- [Composer][composer] **1.x** and **2.x** (Support for Composer v2 is available since [`v0.7.0`][v0.7])
+- [Composer][composer] **2.x**
```
jrfnl, line 51: Is there a reason to remove the historic context remark ?
Might it be more informative to keep it and add info about when support for Composer 1.x was dropped ?

jrfnl (changes requested):
Hi @fredden Thanks for preparing this PR!

I've gone through the codebase with a fine toothcomb and found some more changes which I believe should be made for this PR:
* `CONTRIBUTING.md` line 102 still contains a reference to a `composer1.phar` file.
* `BaseLineTest::testBaseLineLocal()`: I believe the test skip condition can be removed.
* `RemovePluginTest`: I think the `getUninstallStdOutExpectation()` method can be removed and calls to it replaced with `''`.
* `TestCase::writeComposerJsonFile()`: The special casing to disable TLS on Windows with Composer 1.x can be removed.

I look forward to the next iteration of the PR.

@Potherca I wonder if we shouldn't just drop support for Composer < 2.2 (released Dec 2021 vs 2.0 released Oct 2020).
Composer 2.2 is a LTS release with support for PHP < 7.2 and still sees releases, so should definitely stay supported for now.
Also see: https://packagist.org/packages/composer/composer/stats#major/2

What do you think ?

Potherca:
I think that we can drop support for 2.0 and 2.1 as both (combined) have dropped under 1.5K daily downloads out of a +/-65K total, and there aren't many (any?) compelling reasons for people to not update to 2.2 (or higher).

I think <2% is an acceptable cut-off point when perfectly good alternatives exists for the same PHP version(s).

jrfnl:
> I think that we can drop support for 2.0 and 2.1 as both (combined) have dropped under 1.5K daily downloads out of a +/-65K total, and there aren't many (any?) compelling reasons for people to not update to 2.2 (or higher).
> 
> I think <2% is an acceptable cut-off point when perfectly good alternatives exists for the same PHP version(s).

Well, the stats on Packagist are the package downloads. There's also the Phars, Phive and other installation methods, which I suspect in this case to be more prevalent and more relevant, but for which we don't have stats available.

Having said that, I definitely support dropping support for Composer < 2.2.

README.md
fredden: After I removed the "1.x and", the words in parenthesis no longer made sense. I considered rewriting this section into a table; would this be preferable?

composer.json
fredden: What would you like this to read? It's currently `*`, which means "any version".

README.md
fredden: Perhaps a table like this:

|Plugin version|PHP|Composer|PHP_CodeSniffer|
|-|-|-|-|
|`^0.1.0`|`*`|`^1.0`|`*`|
|`^0.2.0`|`*`|`^1.0`|`*`|
|`^0.3.0`|`*`|`^1.0`|`*`|
|`^0.4.0`|`*`|`^1.0`|`*`|
|`^0.5.0`|`^5.3 \|\| ^7.0`|`^1.0`|`^2.0 \|\| ^3.0`|
|`^0.6.0`|`^5.3 \|\| ^7.0`|`^1.0`|`^2.0 \|\| ^3.0`|
|`^0.7.0`|`>=5.3`|`^1.0 \|\| ^2.0`|`^2.0 \|\| ^3.0`|
|`^1.0`|`>=5.4`|`^1.0 \|\| ^2.0`|`^2.0 \|\| ^3.1 \|\| ^4.0`|
|`^1.1`|`>=5.4`|`^2.0`|`^2.0 \|\| ^3.1 \|\| ^4.0`|

fredden:
> `CONTRIBUTING.md` line 102 still contains a reference to a `composer1.phar` file.

I did see this and chose to keep it. Would you prefer that this is removed?

> `BaseLineTest::testBaseLineLocal()`: I believe the test skip condition can be removed.
> `RemovePluginTest`: I think the `getUninstallStdOutExpectation()` method can be removed and calls to it replaced with ''.
> `TestCase::writeComposerJsonFile()`: The special casing to disable TLS on Windows with Composer 1.x can be removed.

I have applied these changes. Initially I chose to not make any PHP changes, but it makes sense to put the obvious ones in this pull request.

composer.json
jrfnl: Well, if we're dropping support for Composer < 2.0, it should become `^2.0` I'd say... (or `^2.2` if we'd go for a Composer 2.2 minimum).

README.md
jrfnl: I don't think a table is needed. That's for Composer to sort out on install. Just some generic info should be enough.

jrfnl:
> > `CONTRIBUTING.md` line 102 still contains a reference to a `composer1.phar` file.
> 
> I did see this and chose to keep it. Would you prefer that this is removed?

Why yes. as the sentence is about running the tests locally and that won't work with a `composer1.phar` anymore, so the information in the CONTRIBUTING guide would be misleading for new contributors.

> 
> > `BaseLineTest::testBaseLineLocal()`: I believe the test skip condition can be removed.
> > `RemovePluginTest`: I think the `getUninstallStdOutExpectation()` method can be removed and calls to it replaced with ''.
> > `TestCase::writeComposerJsonFile()`: The special casing to disable TLS on Windows with Composer 1.x can be removed.
> 
> I have applied these changes. Initially I chose to not make any PHP changes, but it makes sense to put the obvious ones in this pull request.

Curious why you initially chose not to make those PHP changes. Care to explain ?
In my view, in that case, it would be a "drop support in name only", which is not the intention.

README.md
fredden: > some generic info should be enough.

So, should we leave this as-is, or do you want to supply some different text?

jrfnl: What about something like:
`(Support for Composer v2 is available since [`v0.7.0`][v0.7]; Support for Composer < 2.0. was dropped in [`v1.1.0`][v1.1])`

@Potherca What do you think ?

composer.json
```diff
@@ -28,13 +28,13 @@
...
   "require": {
     "php": ">=5.4",
-    "composer-plugin-api": "^1.0 || ^2.0",
+    "composer-plugin-api": "^2.2",
```
jrfnl, line 31: For the record - as I had to double-check this myself:

**The plugin API _does_ follow the same _major_ version as Composer has, but does NOT necessarily align to the same _minors_.**

The current version of the `composer-plugin-api` is `2.6.0` while the Composer package itself is at `2.8.4`. See:
https://github.com/composer/composer/blob/f5e7a8390d700d09c87ae2ed79d5aea1e4153cd3/src/Composer/Plugin/PluginInterface.php#L35

However, in this specific case, the versions happen to line up for the `2.2.0` version of Composer:
https://github.com/composer/composer/blob/e174a4c4324f50a6f2de472aa1055c24a2fe2b2a/src/Composer/Plugin/PluginInterface.php#L35

CONTRIBUTING.md
```diff
@@ -99,7 +99,7 @@ In case the test setup has trouble locating your `composer.phar` file:
     </php>
     ```
     **Note**: this setting also allows for locally testing with different versions of Composer.
-    You could, for instance, have multiple Composer Phar files locally, `composer1.phar`, `composer2.1.phar`, `composer2.2.phar`.
+    You could, for instance, have multiple Composer Phar files locally, `composer.phar`, `composer2.1.phar`, `composer2.2.phar`.
```
jrfnl, line 102:
```suggestion
    You could, for instance, have multiple Composer Phar files locally, `composer.phar`, `composer2.2.phar`, `composer2.6.phar`.
```

If we're dropping support for Composer < 2.2, mentioning a `2.1` phar is also confusing.

README.md
```diff
@@ -48,7 +48,7 @@ That's it.
 This plugin is compatible with:
 
 - PHP **5.4+**, **7.x**, and **8.x** (Support for PHP v8 is available since [`v0.7.0`][v0.7])
-- [Composer][composer] **1.x** and **2.x** (Support for Composer v2 is available since [`v0.7.0`][v0.7])
+- [Composer][composer] **2.2+** (Support for Composer v2 is available since [v0.7.0][v0.7]; support for Composer < 2.2 was dropped in [v1.1.0][v1.1])
```
jrfnl, line 51:
```suggestion
- [Composer][composer] **2.2+** (Support for Composer v2 is available since [`v0.7.0`][v0.7]; support for Composer < 2.2 was dropped in [`v1.1.0`][v1.1])
```

README.md
jrfnl, line 1: As we seem to be dropping support for Composer 2.0 < 2.2 in this same PR now, the text about the permissions change in Composer 2.2. reads a bit dated.

What about updating line 26 like so:
```diff
-When using Composer 2.2 or higher, Composer will...
+Since Composer 2.2, Composer will ...
```
and then updating line 39 like so:
```diff
-When using Composer < 2.2, you can add ...
+When using an older version of the plugin in combination with Composer < 2.2, you can ...
```

jrfnl (changes requested):
Thanks for these updates @fredden !

The PR now also includes the version drop for Composer 2.0 < 2.2, which means some additional changes are needed. Most have been made, but I've left some inline comments related to this + I think the `InstallUpdateEventsTest::testPluginRunsOnInstallWithNoScripts()` test can get some clean up for the Composer 2.0 < 2.2 version drop.

CONTRIBUTING.md
fredden: Good catch. I'll update these version numbers too.

README.md
fredden: I've reworded these paragraphs.

README.md
fredden: :+1:

tests/IntegrationTest/InstallUpdateEventsTest.php
```diff
@@ -177,7 +177,7 @@ public function testPluginRunsOnReinstall()
     }
 
     /**
-     * Test that the plugin runs (or doesn't run) when Composer is invoked with the --no-scripts argument.
+     * Test that the plugin runs when Composer is invoked with the --no-scripts argument.
      *
      * Note: the behaviour of Composer changed in 2.1.2. Prior to that, `--no-scripts` would
```
jrfnl, line 182: I think this "Note" can now be removed as well (as it's no longer relevant).

fredden: I kept this because it explains why the test exists. Should we instead drop the whole test?

jrfnl: Understood and no, let's keep the test as it has value either way.

jrfnl (approved):

Potherca (approved):

jrfnl:
FYI: I've updated the branch protection to be in line with this change.

@potherca Do we want to tag a release for this already ? Or should we include the "drop support for PHPCS 2.x" task into the same minor ?

Potherca:
I'm in favor of keeping this as part of 1.1

jrfnl:
@Potherca Sorry, there appears to be confusion. Let me rephrase the question a little:

Do we want to tag the `1.1.0` release now this PR has been merged ? Or do we want to include the "drop support for PHPCS 2.x" task into the same minor ? (otherwise that one would move to the `1.2.0` release (or later))

Potherca:
I don't really have a rationale but my feeling/instinct is to split the releases, 1.1 now and 1.2 with the Composer <2.2 changes.

jrfnl:
@Potherca Same intuition I had, which is why I asked the question ;-)

Though to be clear, this would mean:
* `1.1.0`: drop support for Composer < 2.2
* `1.2.0`: drop support for PHPCS 2.x
* `1.?.x`: drop support for PHP < 7.2.

Agreed ? How do you want to do the release ? Quick videocall with the two of us some time next week ?

fredden:
@Potherca please can you tag/release version `1.1.0`. I can then get started with the proposed changes for `1.2.0` as above.

PR #234, opened 2025-04-02
Plugin: use Composer API 2.x methods

jrfnl:
### Modernize: use InstalledVersions::isInstalled() 

The Composer runtime-api 2.0 introduced the `\Composer\InstalledVersions::isInstalled()` method.

This commit starts using that method.

Implementation choices:
While it would be possible to also handle the version checking via `Composer\InstalledVersions::satisfies(new VersionParser, self::PACKAGE_NAME, (string) $versionConstraint)`, this would require adding a new dependency `"composer/semver": "^3.0"`.
For now, I've chosen not to add the dependency and to leave the pre-existing code for handling version constraints.

Ref: https://getcomposer.org/doc/07-runtime.md#installed-versions

### Modernize: use InstalledVersions::getInstallPath() 

The Composer runtime-api 2.1 introduced the `\Composer\InstalledVersions::getInstallPath()` method.

This commit starts using that method.

Ref: https://getcomposer.org/doc/07-runtime.md#knowing-the-path-in-which-a-package-is-installed

## Related Issues

Follow up on #230

👉🏻 Note: I have another modernization lined up, but am running into a bug in Composer. Holding that one back for now.

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -12,6 +12,7 @@ namespace PHPCSStandards\Composer\Plugin\Installers\PHPCodeSniffer;
 
 use Composer\Composer;
 use Composer\EventDispatcher\EventSubscriberInterface;
+use Composer\InstalledVersions;
 use Composer\IO\IOInterface;
 use Composer\Package\AliasPackage;
 use Composer\Package\PackageInterface;
@@ -566,6 +567,10 @@ class Plugin implements PluginInterface, EventSubscriberInterface
      */
     private function isPHPCodeSnifferInstalled($versionConstraint = null)
     {
+        if ($versionConstraint === null) {
+            return InstalledVersions::isInstalled(self::PACKAGE_NAME);
+        }
+
         return ($this->getPHPCodeSnifferPackage($versionConstraint) !== null);
     }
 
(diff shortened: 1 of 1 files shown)


jrfnl:
Hang on - this PR still contained a commit from another PR, so I've rebased this to drop that extraneous commit (without other changes)

fredden (approved):

Potherca (approved):
Looks good. Just to be clear: We're breaking backward compatibility on purpose, right? 

I.e, we're not checking whether `Composer\InstalledVersions` exist, as we are bumping the required version of composer to versions that are quaranteed to contain `Composer\InstalledVersions`.

(Just double-checking, as I've been out of touch with this project for a while).

jrfnl:
> Looks good. Just to be clear: We're breaking backward compatibility on purpose, right?
> 
> I.e, we're not checking whether `Composer\InstalledVersions` exist, as we are bumping the required version of composer to versions that are quaranteed to contain `Composer\InstalledVersions`.

Exactly, the [version bump for the Composer API](https://github.com/PHPCSStandards/composer-installer/pull/230/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L31) was already done in PR #230 and as this tool is a Composer plugin and therefore only needs to take installation via Composer into account, we don't need to check with `function_exists()` or anything.
IF the version of this package is installed containing this change, this means by definition that the user WILL have the right Composer version installed. (Otherwise, they wouldn't get the new version of this package, as Composer would have installed the last version from before the drop of support for Composer < 2.2).

PR #235, opened 2025-05-02
Tests: run against the *new* PHPCS 4 branch

jrfnl:
## Proposed Changes

The PHP_CodeSniffer 4.0 branch has been completely rebuild. See https://github.com/PHPCSStandards/PHP_CodeSniffer/issues/120 for more detailed information about the "why".

In practice, this means that to test against the upcoming PHPCS 4.0 release, we need to test against the `4.x` branch now, not the old `4.0` branch.

A 4.0 beta/RC release is expected in the next few days. Once that is out, a test period starts and depending on whether problems are found or not, the 4.0.0 release will be tagged about a month later (or later if significant problems were found).

When 4.0.0 gets tagged, the PHPCS `master` branch will be renamed to `3.x` and the `4.x` branch will be the new "main" branch for PHPCS, so more changes will be needed, but that's for later.

Includes updating the `DemoSniff` test fixture setup as PHPCS 4.0 will start enforcing for sniffs to implement the `Sniff` interface.

### DIFF
--- a/tests/PHPCSVersions.php
+++ b/tests/PHPCSVersions.php
@@ -27,7 +27,7 @@ final class PHPCSVersions
      *
      * @var string
      */
-    const NEXT_MAJOR = '4.0.x-dev as 3.9.99';
+    const NEXT_MAJOR = '4.x-dev as 3.99.99';
 
     /**
      * List of all PHPCS version which are supported by this plugin.
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -30,6 +30,7 @@
 
     <rule ref="PSR1.Files.SideEffects">
         <exclude-pattern>*/tests/bootstrap\.php$</exclude-pattern>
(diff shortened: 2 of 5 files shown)


tests/PHPCSVersions.php
```diff
@@ -27,7 +27,7 @@ final class PHPCSVersions
      *
      * @var string
      */
-    const NEXT_MAJOR = '4.0.x-dev as 3.9.99';
+    const NEXT_MAJOR = '4.x-dev as 3.9.99';
```
fredden, line 30: Version 3.10 already exists.
```suggestion
    const NEXT_MAJOR = '4.x-dev as 3.99.99';
```

fredden (approved):

tests/PHPCSVersions.php
jrfnl: Ouch.. thanks for catching that and yes, 100% agree, that's what it should have read in the first place.

fredden (approved):

Potherca (approved):

PR #244, opened 2025-06-25
Move release-checklist to separate file & update it

jrfnl:
## Proposed Changes

The updated release checklist presumes the changes in PR #243 will be accepted.

The release checklist follows a similar format as other release checklists in this organisation.


Potherca (commented):
Two nitpicks: I would suggest using actual emoji rather than emoji shortcodes, and either sub-list or `<br>` for newlines. 

The rational for both is they make the markdown more likely to correctly/consistenly render across environments.

jrfnl:
@Potherca Fair points. Should be fixed now. Mind: I've moved this PR to the _next_ milestone to prevent it blocking the release.

PR #245, opened 2025-06-25
Fix "The provided cwd "../squizlabs/php_codesniffer" does not exist."

### LINKED ISSUE #239: Composer update in a drupal stack will break after composer-installer new version
## Describe the bug

After https://github.com/PHPCSStandards/composer-installer/releases/tag/v1.1.0 new release, in a drupal stack, with https://github.com/drupal/core-dev, which along its children dependencies has composer-installer, and php_codesniffer, and a `composer update` when updating PHPCSStandards/composer-installer version from 1.0.0 to 1.1.0 will break with the following error:

```
[38.1MiB/36.16s] Scaffolding files for drupal/core:

[..snip..]

[38.1MiB/36.16s]   - Copy [web-root]/sites/default/default.settings.php from assets/scaffold/files/default.settings.php

[38.2MiB/36.17s] ::error ::The provided cwd "../squizlabs/php_codesniffer" does not exist.
In Process.php line 350:
                                                                   
  The provided cwd "../squizlabs/php_codesniffer" does not exist.  
                                                                   
update [--with WITH] [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--dry-run] [--dev] [--no-dev] [--lock] [--no-install] [--no-audit] [--audit-format AUDIT-FORMAT] [--no-autoloader] [--no-suggest] [--no-progress] [-w|--with-dependencies] [-W|--with-all-dependencies] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [-m|--minimal-changes] [--patch-only] [-i|--interactive] [--root-reqs] [--bump-after-update [BUMP-AFTER-UPDATE]] [--] [<packages>...]
```

Still with `drupal/core-dev` in my composer.json, and i enforce composer-installer to the previous version, error is gone.

```
"dealerdirect/phpcodesniffer-composer-installer": "v1.0.0"
```

Not sure if this would be a PHPCSStandards/PHP_CodeSniffer issue, or PHPCSStandards/composer-installer.

## Please confirm

- [ ] I have searched the issue list and am not opening a duplicate issue.
- [ ] I have read the [Contribution Guidelines](https://github.com/PHPCSStandards/PHP_CodeSniffer/blob/master/.github/CONTRIBUTING.md) and this is not a [support question](https://github.com/PHPCSStandards/PHP_CodeSniffer/discussions).
- [ ] I confirm that this bug is a bug in PHP_CodeSniffer and not in one of the external standards.
- [ ] I have verified the issue still exists in the `master` branch of PHP_CodeSniffer.

jrfnl:
## Proposed Changes

I'm very curious why the reported issue only appears to happen with a Drupal setup (and I cannot reproduce it, nor do our tests fail). All the same, I'm hoping this will fix it.

Note: yes, `realpath()` can return `false`, but that shouldn't be possible in this case as we only call it after we've already checked `isPHPCodeSnifferInstalled()`.

## Related Issues

Fixed #239 (hopefully)

### DIFF
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -555,7 +555,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
      */
     private function getPHPCodeSnifferInstallPath()
     {
-        return (string) InstalledVersions::getInstallPath(self::PACKAGE_NAME);
+        return (string) realpath(InstalledVersions::getInstallPath(self::PACKAGE_NAME));
     }
 
     /**
--- a/src/Plugin.php
+++ b/src/Plugin.php
@@ -555,7 +555,7 @@ class Plugin implements PluginInterface, EventSubscriberInterface
      */
     private function getPHPCodeSnifferInstallPath()
     {
-        return (string) realpath(InstalledVersions::getInstallPath(self::PACKAGE_NAME));
+        return $this->composer->getInstallationManager()->getInstallPath($this->getPHPCodeSnifferPackage());
     }
 
     /**


jrfnl:
Thanks for testing @rpkoller ! That sounds like `realpath()` returning `false`, so something must be off about the Composer API method as that shouldn't be possible.

I've just pushed a new commit to revert this line back to its v1.0 state. Would be great to get confirmation whether that does fix things.

rpkoller:
with the change to `return $this->composer->getInstallationManager()->getInstallPath($this->getPHPCodeSnifferPackage());` a composer require runs through without an error again

sharif-elshobkshy:
@jrfnl  . Thank you for your review.

That seems to fix the bug. 

```
$ composer update
Gathering patches for root package.
Loading composer repositories with package information
Updating dependencies
Nothing to modify in lock file
Installing dependencies from lock file (including require-dev)
Nothing to install, update or remove
Package laminas/laminas-text is abandoned, you should avoid using it. No replacement was suggested.
Generating autoload files
Hardening vendor directory with .htaccess and web.config files.
102 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
PHP CodeSniffer Config installed_paths set to ../../drupal/coder/coder_sniffer,../../sirbrillig/phpcs-variable-analysis,../../slevomat/coding-standard
phpstan/extension-installer: Extensions installed
Cleaning installed packages.
```

pandaskii:
@jrfnl

This new change resolves the issue.

I was troubleshooting it in the resolved install paths, and here are the results:

`/app/vendor/squizlabs/php_codesniffer` is the correct and expected path

```
DEBUG: loadInstalledPaths called
DEBUG: InstalledVersions path 1: /app/vendor/composer/../squizlabs/php_codesniffer (the new method from 1.1.0)
DEBUG: InstalledVersions path 2: /app/vendor/squizlabs/php_codesniffer
```

To debug this, I modified the following code accordingly.

```
/**
     * Returns the path to the PHP_CodeSniffer package installation location
     *
     * @return string
     */
    private function getPHPCodeSnifferInstallPath()
    {
        error_log("DEBUG: loadInstalledPaths called");
        $path1 = (string) InstalledVersions::getInstallPath(self::PACKAGE_NAME);
        $path2 = $this->composer->getInstallationManager()->getInstallPath($this->getPHPCodeSnifferPackage());
        error_log("DEBUG: InstalledVersions path 1: " . $path1);
        error_log("DEBUG: InstalledVersions path 2: " . $path2);
        return $path2;
    }
```

MarcHagen:
Side note, not only Drupal setups are affected. My personal dev pc and multiple action workflows are failing with exactly the same error.

jrfnl:
@MarcHagen Could you please add details about your setup to #239 ? We still haven't been able to pinpoint the underlying cause, so would be great to get more information to see if we can find a pattern.

PR #250, opened 2025-07-09
Tests: fix two missing import `use` statements

jrfnl:
## Proposed Changes

Add two import `use` statements for the PHP native `RuntimeException` class, which is used to signal certain failures in the test set up.

The fact that this wasn't noticed before clearly means the tests never failed at those Exceptions before ;-)

## Suggested changelog entry
_N/A_ (general housekeeping)


Potherca (approved):

jrfnl:
> I have no issues with approving these changes, but as I recall, the imports _not_ being there _shouldn't_ cause an issue.
> 
> Not finding `RuntimeException` in `PHPCSStandards\Composer\Plugin\Installers\PHPCodeSniffer\Tests\IntegrationTest`, PHP should traverse the namespace until it reaches "" and use the global RuntimeException, right?
> 
> (or am I mixing up languages again? 😕 )

Yes, you're mixing up languages.

In namespaced files, there is no fallback to the global namespace (or traversing) for classes. Only functions and constants fall back to the global namespace.

So, no, this missing `use` statement results in a fatal "class not found" error (which is how I came across it).

PR #259, opened 2025-09-30
README: update examples

jrfnl:
## Proposed Changes

The [Object Calisthenics](https://github.com/object-calisthenics/phpcs-calisthenics-rules) project has been deprecated/abandoned, so I don't think mentioning it in the example is a good idea.

Aside from that, both PHPCompatibility and WordPressCS now have new dependencies which register additional standards with PHPCS, so the example output was outdated too.

Lastly, let's update the PHPCS version used in another example to a more recent one.

## Suggested changelog entry
* Updated documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,8 @@ After running `composer install` PHP_CodeSniffer just works:
 
 ```bash
 $ ./vendor/bin/phpcs -i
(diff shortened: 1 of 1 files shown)


Potherca (dismissed):

jrfnl:
Noticed one small mistake in this PR, which I've just fixed up. PR should be squash-merged once the build has run.

PR #260, opened 2025-09-30
Tests/PHPCSVersions: improve handling PHPCS 4.x

jrfnl:
## Proposed Changes

Follow up on #255, which missed making this change.

While not that relevant yet as the only PHPCS 4.x release is currently 4.0.0, this will become relevant once additional PHPCS 4.x releases have been tagged.

## Suggested changelog entry
_N/A_ (test only change)


jrfnl:

PR #261, opened 2025-09-30
Drop support for PHPCS 2.x

jrfnl:
## Proposed Changes

This commit drops support for PHPCS 2.x as discussed in #221.

There may be more tweaks to be made, but this should get rid of most of the dead wood.

## Suggested changelog entry
Support for PHP_CodeSniffer 2.x has been dropped.


jrfnl:
In light of the approval of this PR, I have re-milestoned everything which was milestoned for 1.1.x to the 1.2.0 milestone.

jrfnl:

jrfnl:
@Potherca Do we want to have this PHPCS 2.x version drop in a separate minor ? I.e. should the PHP < 7.2 version drop go into this release (`1.2.0`) or into the next minor (`1.3.0`) ?

jrfnl:
> I'm usually in favor of deprecating or dropping things one at a time, i.e. one per release. Unless there are pressing reasons. (FYI: Lessening the workload of maintainers can count as "pressing").

I tend to agree, so I guess that means we should release 1.2.0 some time over the next few days/weeks ?

As a side-note: reason for me to prioritize the version drops at the moment, is, in part, to make the surface for those Windows errors smaller, which should help with tracking down what's going on there.

Just doing a plain `composer install` with a `composer.json` containing PHPCS, this plugin and a few external standards works fine on Windows (10), even with PHP 8.5, so this really needs a deep-dive to:
1. Try to reproduce the hanging of the test suite;
2. Figure out the exact test on which the test suite hangs
3. Figure out what's going on with that test.
... and as running the tests on Windows is very slow, this is pretty darn time consuming.

Even though the plugin itself is small, the matrix of all combinations of dependencies (Composer versions, PHPCS versions, PHP versions, OSes and even PHPUnit versions) make it hard to figure out what's going on, so anything that limits that matrix of dependencies will help debug the Windows issue - and, as I suspect this is not an issue with the plugin itself, will hopefully help figure out in which dependency the problem is.
Once we know that, we can figure out whether it is a case of excluding a certain dependency version (`conflicts`) or reporting/fixing the issue upstream if the issue still exists in the latest version of the dependency.

PR #262, opened 2025-10-04
Tests/PHPCSVersions: bug fix 

jrfnl:
### Tests/PHPCSVersions: bug fix 

The PHP native `min()` and `max()` versions were used to retrieve the lowest/highest PHPCS version from arrays of PHPCS versions.

This worked fine at the time this class was created as none of the supported PHPCS versions had a `*.10` minor (or higher) at the time.

However, this logic was broken as soon as the `3.10.0` release was tagged - see https://3v4l.org/XbHOY - , we just never realized until now.

Fixed now by adding custom `min`/`max` methods to the class.

---

Note: the first two commits are just girl-scouting while I was looking at the file anyway. The third commit is why I was looking at the file ;-)

## Suggested changelog entry
_N/A_ (test only change)

### DIFF
--- a/tests/PHPCSVersions.php
+++ b/tests/PHPCSVersions.php
@@ -153,8 +153,8 @@ final class PHPCSVersions
         $selection = array();
 
         if (empty($versions) === false) {
-            $selection[] = min($versions);
-            $selection[] = max($versions);
+            $selection[] = self::min($versions);
+            $selection[] = self::max($versions);
         }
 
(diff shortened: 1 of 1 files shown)


jrfnl:
> I could say something about code duplication, but as this is test-code, I think the duplication is warranted.

Fair. I just focused on getting it fixed for now. We can always make it prettier later ;-)

PR #280, opened 2026-03-07
Yamllint: update config to fix (ignore) warnings

jrfnl:
## Proposed Changes

Yamllint 1.37.1 introduced a new check which check the number of spaces between the Yaml and trailing comments on the same line.

In their wisdom, they have elected to make the expected number of spaces default to `2`, while we use `1` in practice.

This commit updates the configuration to check for 1 space instead of 2.

Ref:
* https://github.com/adrienverge/yamllint/blob/master/CHANGELOG.rst#1371-2025-05-04
* https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.comments

## Suggested changelog entry
_N/A_

### DIFF
--- a/.github/workflows/label-merge-conflicts.yml
+++ b/.github/workflows/label-merge-conflicts.yml
@@ -1,3 +1,4 @@
+---
 name: Check PRs for merge conflicts
 
 on:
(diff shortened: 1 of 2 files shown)


jrfnl:
Just in case anyone is wondering - the warnings I'm talking about can be seen here: https://github.com/PHPCSStandards/composer-installer/actions/runs/25436575221

Potherca (approved):

Potherca (approved):
Yeah, I just encountered the wall of Yaml warnings. 😿 

Merge at will.

PR #281, opened 2026-03-07
Drop support for PHP < 7.2

### LINKED ISSUE #221: Proposal for limiting supported versions

jrfnl:
### Drop support for PHP < 7.2 

This commit comprehensively drops support for PHP < 7.2.

Includes changing the security workflow to use the Composer `audit` command instead of the deprecated security checker CLI tool.

**Note:** this commit does not change the supported PHPUnit versions (other than no longer needing PHPUnit 4.x - 7.x).
This could possibly be done in a follow-up ticket, but to support PHPUnit >= 10, we'll need to find another way to debug failing tests as the test listener implementation we currently use, is no longer supported on PHPUnit >= 10.

Closes #221

### Modernize: add constant visibility 

For the tests, all visibility modifiers have been set to `private`.
For the `src`, all visibility modifiers have been set to `public` as anything else could be considered a BC-break.

I suggest we open a ticket to change the constant visibility for the class constants in the `Plugin` class to `private` in the next major.

## Suggested changelog entry
Removed: Support for PHP < 7.2

### DIFF
--- a/.github/workflows/integrationtest.yml
+++ b/.github/workflows/integrationtest.yml
@@ -26,11 +26,6 @@ jobs:
     strategy:
       matrix:
         php:
-          - '5.4'
-          - '5.5'
-          - '5.6'
-          - '7.0'
-          - '7.1'
           - '7.2'
           - '7.3'
           - '7.4'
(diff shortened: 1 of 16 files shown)


fredden (approved):

jrfnl:
@Potherca Now the 1.2.1 release is out of the door, when you have some time, would be great to get this PR and the other two PRs reviewed, merged and released as v1.3.0. 🙏🏻 

Note: the build for this PR will not pass as the branch protection settings will need to be updated, but it only makes sense to do this when we're ready to merge this PR (so as not to block other PRs).

Potherca (approved):
👍

PR #292, opened 2026-09-07
Tests/PHPCSVersions: update for new PHPCS releases

jrfnl:
## Proposed Changes

Update for new PHPCS releases.

Refs:
* https://github.com/PHPCSStandards/PHP_CodeSniffer/releases/

## Suggested changelog entry
_N/A_


tests/PHPCSVersions.php
```diff
@@ -90,8 +90,11 @@ final class PHPCSVersions
         '3.13.3' => '3.13.3',
         '3.13.4' => '3.13.4',
         '3.13.5' => '3.13.5',
+        '3.13.6' => '3.13.6',
         '4.0.0'  => '4.0.0',
         '4.0.1'  => '4.0.1',
+        '4.0.2'  => '4.0.2',
+        '4.0.4'  => '4.0.4',
```
Potherca, line 97: Might want to add a comment so future developers do not accidentally think `4.0.3` is missing (as they might not know about the snafu).

Also, could 4.0.2. not also be left out? Or did I misinterpret https://github.com/PHPCSStandards/PHP_CodeSniffer/releases/tag/4.0.4 ?

Potherca (dismissed):

tests/PHPCSVersions.php
jrfnl: Fair point. I've removed the `4.0.2` release from the list and replaced it with a comment to explain the missing version numbers.

Potherca (approved):

